PR #11, opened 2022-06-16
docs: taxonomy parser reference

### LINKED ISSUE #7: Write a specification for taxonomy parser / renderer
### What

- Specify how the .txt file should be parsed by the taxonomy editor to get target structure.

### Part of
- #5

alexgarel:
Fixes #7

### DIFF
--- a/doc/references/taxonomy-parser.md
+++ b/doc/references/taxonomy-parser.md
@@ -2,7 +2,7 @@
 
 References: https://wiki.openfoodfacts.org/Global_taxonomies#format
 
-This specification takes the imagine that we have a graph tool.
+This specification assumes that we have a graph tool.
 It can be in memory graph like networkx or custom dictionary structure.
 Or a database, be it a real graph database like Arango db, or another database (eg. PostgresSQL) used with graph queries.
 
@@ -13,7 +13,7 @@ You must be able to query the database on it's relations, but also on nodes prop
 ## Lines and blocks
 
 1. a language code is a two character code. We write `LC` to signify a language code in this document.
-1. every line has on of the following form:
+1. every line has one of the following forms:
    - *blank line*: it only contains space characters or no character at all
    - *comment line*: first character is a `#`
    - *parent line*: fist character is `<`
@@ -38,8 +38,8 @@ You must be able to query the database on it's relations, but also on nodes prop
 1. entry lines are composed of the `LC:` header,
    then a list of *tags* separated by a comma `,`.
    Comma without at least one space around are not separators.
-1. each *tag* as a *normalized version* referred to as *tagid*
-1. each *entry line* as a first tag which is referred to as *line tag*,
+1. each *tag* has a *normalized version* which is referred to as *tagid*
+1. each *entry line* has a first tag which is referred to as *line tag*,
    and as *line tagid* for it's *normalized version*.
    *line tagid* is considered the identifier for this *entry line*,
    and must be unique.
@@ -106,7 +106,7 @@ Replace html escaped chars by "-":
 string = re.sub(r"&\w+;", "-", string)
 ```
 
-Replace some weirds chars by "-" [^strech_weird_chars]:
+Replace some strange characters with "-" [^strech_weird_chars]:
 ```
 string = re.sub(r"[\s!"#\$%&'()*+,\/:;<=>?@\[\\\]^_`{\|}~¡¢£¤¥¦§¨©ª«¬®¯°±²³´µ¶·¸¹º»¼½¾¿×ˆ˜–—‘’‚“”„†‡•…‰‹›€™\t]", "-", string)
 ```
--- a/doc/references/taxonomy-parser.md
+++ b/doc/references/taxonomy-parser.md
@@ -93,7 +93,7 @@ to know if following operations applies, based on language code:
 * lowercase
 * removing accents
 
-If lowercase applies and string is not a UUID, that is it does not match regexp `r"^[a-z\-]+\.[a-zA-Z0-9-_]{8}[a-zA-Z0-9-_]+$"`,
+If lowercase applies and string is not a UUID [^uuid], that is it does not match regexp `r"^[a-z\-]+\.[a-zA-Z0-9-_]{8}[a-zA-Z0-9-_]+$"`,
 then lowercase it. And also transform "." to "-".
 
 Replace every special character (ascii 0 to 27) and zero width space with "-":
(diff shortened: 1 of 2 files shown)


doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
...
+       should be considered a separated block,
+       (with preceding comment lines)
+    1. the first lines, up to the first *meaningful line*
+       (but preceding comment lines) is the *header*
+    1. other blocks must contains at least one *entry line*.
+       We call them *entry blocks*
+1. entry lines are composed of the `LC:` header,
+   then a list of *tags* separated by a comma `,`.
+   Comma without at least one space around are not separators.
+1. each *tag* as a *normalized version* referred to as *tagid*
```
aadarsh-ram, line 41:
```suggestion
1. each *tag* as a *normalized version* is referred to as *tagid*
```

doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
...
+1. each *entry line* as a first tag which is referred to as *line tag*,
```
aadarsh-ram, line 42:
```suggestion
1. each *entry line* has a first tag which is referred to as *line tag*,
```

doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
...
+we will apply that to tag to obtain the tagid.
+
+It depends on the language.
+
+Obtain NFC form of the string. In python:
+```python
+string = unicodedata.normalize("NFC", my_input)
+```
+
+Then look at `string_normalization_for_lang` in `Config_off.pm` [^string_config]
```
aadarsh-ram, line 91: A link to the Perl module and function would be appreciated. 
A separate .md file for the language codes would be easier for reference.

doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
...
+to know if following operations applies, based on language code:
+* lowercase
+* removing accents
+
+If lowercase applies and string is not a UUID, that is it does not match regexp `r"^[a-z\-]+\.[a-zA-Z0-9-_]{8}[a-zA-Z0-9-_]+$"`,
```
aadarsh-ram, line 96: Would you be able to elaborate on what a *UUID* is, with reference to this document?

doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
...
+```python
+string = re.sub(r"[\u0000-\u0027\u200b]", "-", string)
+```
+
+Replace html escaped chars by "-":
+```python
+string = re.sub(r"&\w+;", "-", string)
+```
+
+Replace some weirds chars by "-" [^strech_weird_chars]:
```
aadarsh-ram, line 109:
```suggestion
Replace some strange characters by "-" [^strech_weird_chars]:
```

doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
...
+```python
+string = re.sub(r"-+", "-", string)
+string = string.strip("-")
+```
+
+[^string_config]: of course you would have copied it to a local config file, or a dict in your code.
+
+[^strech_weird_chars]: (stretch goal: use [unicode category](https://www.compart.com/en/unicode/category) (through unicodedata module) to remove all unwanted category)
+
+# Processing
```
aadarsh-ram, line 126: Please add the **current logic/process** followed during *taxonomy validation* in this section. This would be helpful for implementation in the future.

aadarsh-ram (changes requested):

doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
...
+   * every *parent tag* with their *tagid*, language code included (this is temporary)
+   * every *tags*, with their *tagid*, language code included
+
+For every node, also store:
+* a specific `block_index` property which stores the number of blocks before this one in the file
+* a specific `block_subindex` property which is always 0
+  (such that if we want to insert a new block beneath this block,
+  later on,
+  we could insert it with same block_index but block_subindex of 1)
+* preceding *comments lines* in a `preceeding_lines` property
```
aadarsh-ram, line 147: As far as I can understand, the *preceeding_lines* id stores the number of comment lines present before the current block. But, a way for storing comments isn't mentioned in the document. 

Would it be optimal to add a tag called *comment*, with its value as the comment line?

doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
+# Taxonomy parser
+
+References: https://wiki.openfoodfacts.org/Global_taxonomies#format
+
+This specification takes the imagine that we have a graph tool.
```
aadarsh-ram, line 5:
```suggestion
This specification assumes that we have a graph tool.
```

doc/references/taxonomy-parser.md
```diff
@@ -0,0 +1,161 @@
...
+Or a database, be it a real graph database like Arango db, or another database (eg. PostgresSQL) used with graph queries.
+
+You must be able to query the database on it's relations, but also on nodes properties in an efficient way.
+
+# Definitions
+
+## Lines and blocks
+
+1. a language code is a two character code. We write `LC` to signify a language code in this document.
+1. every line has on of the following form:
```
aadarsh-ram, line 16:
```suggestion
1. every line has one of the following forms:
```

doc/references/taxonomy-parser.md
alexgarel:
```suggestion
1. each *tag* has a *normalized version* which is referred to as *tagid*
```

doc/references/taxonomy-parser.md
aadarsh-ram: @alexgarel Regarding "validation of a taxonomy", there should be a *logic/code* in place currently. Adding that logic to this document, would be helpful for implementation.

alexgarel: hum @stephanegigandet might tell me if I'm wrong, but I think we only verify we don't have same term twice in different entries (and currently we do not even enforce it, for we know some taxonomies have this problem).

I think we can leave it to the editor. But we should write a spec about it, you're right, but not at parsing time.

alexgarel:
@aadarsh-ram can you validate, so that we merge, it's not perfect, but we can improve there after.

aadarsh-ram (approved):

PR #17, opened 2022-07-05
build: add neo4j with docker

alexgarel:
Add a neo4j database using docker-compose (basic setup for now).


BryanH01 (approved):
It seems to be working

PR #18, opened 2022-07-06
feat: taxonomy parser library

BryanH01:
### What
- Created parser python script in parser folder
- It's currently only harvesting data and creating the nodes

I'm not sure about the use of a class here.

### Related issue(s)
Fixes #22


parser/Parser.py
```diff
@@ -0,0 +1,133 @@
...
+                for word in line[3:].split(","):
+                    new_line+=normalizing(word,lang)+','
+                new_line=new_line[:-1]
+                return new_line
+
+        index_comm=0
+        index_stopwords=1
+        index_synonyms=1
+        stop = re.compile('stopword')
+        syno = re.compile('synonym')
```
alexgarel, line 63: avoid too short variable name. It makes code less readable.

Here, maybe : synonym_prefix.

You don't need a regexp here, do you ?

BryanH01: yes, you're right, i don't need them

parser/Parser.py
```diff
@@ -0,0 +1,133 @@
...
+                    file.append([])
+            else:
+                #sanitizing the line before adding it
+                line = line.rstrip()
+                line = line.replace("’", "'")
+                line = re.sub(r"(\d),(\d)", r"\1‚\2", line)
+                line = re.sub(r"\\,", "\\‚", line)
+                line = re.sub(r"\(([ivx]+)\)", r"\1", line, flags=re.I)
+                file[-1].append(line)
+        self.file = file
```
alexgarel, line 29: Here you could write a generator:
```python
    def file_iter(self)
        for line in txtfile:
            # do sanitize
            yield line
        
        
   def harvest(self):
       for line in self.file_iter():
           ... process line
```

parser/Parser.py
```diff
@@ -0,0 +1,133 @@
...
+        lc = re.compile('[a-zA-Z][a-zA-Z]:')
+
+        #header
+        h=0
+        header=''
+        while self.file[h][0][0]=="#":
+            for line in self.file[h]:
+                header+=line
+            h+=1
+        query="CREATE (node{id:'__header__'" + (",value:"+header if len(header)>0 else "") +"})"
```
alexgarel, line 73: it would be better to have sub method for writing to database, it will really ease unit testing but also it's cleaner.

parser/Parser.py
BryanH01: Oh i didn't know this

parser/Parser/Parser.py
```diff
@@ -0,0 +1,174 @@
...
+
+class Parser:
+
+    def __init__(self,filename,uri="bolt://localhost:7687"):
+        self.driver = GraphDatabase.driver(uri)
+        self.session = self.driver.session() #Doesn't create error even if there is no active database
+        self.filename=filename
+
+    def file_name(self):
+        """open the file filename and return the list of entries, each entry is a list a the sanitized lines"""
```
alexgarel, line 13: update documentation ;-)

parser/Parser/Parser.py
```diff
@@ -0,0 +1,174 @@
...
+        self.filename = self.filename + ( '.txt' if (len(self.filename)<4 or self.filename[-4:]!=".txt") else '')  #to not get error if extension is missing
+
+    def file_iter(self,start=0):
+        """generator to get the file line by line"""
+        counter=0
+        with open(self.filename,"r",encoding="utf8") as file:
+            for line in file:
+                if counter < start : counter+=1
```
alexgarel, line 21: for this kind of check it's ok to use a "continue" statement and avoid the else:

```
if counter < start:
    counter += 1
    continue
```

even better you can use itertools.islice:

```
for line in itertools.islice(file, start=start):
```

parser/Parser/Parser.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def create_node(self,data):
+        id_query = " CREATE (n{id:'" + data['id'] + "'}) SET n.block_index=" + str(self.block_index) + ", n.block_subindex=0, n.preceding_lines="+str(data['comment'])
+        if data['value'] : entry_query = " SET n.value='" + str(data['value']) + "'"
+        else : entry_query = " SET n.tags=\"" + str(data['tags']) + "\" ,n.tagsid=" + str(data['tagsid']) + " ,n.parents=" + str(data['parent_tag']) + " ,n.properties=" + str(data['properties'])
+        query=id_query+entry_query
+        # print(query)
+        self.session.run(query)
+
+    def reinitialize_data(self):
```
alexgarel, line 74: I would named it: `new_node_data` or `initial_data` (it's not a transformation you apply to something)

parser/Parser/Parser.py
```diff
@@ -0,0 +1,174 @@
...
+            parent_list=result["n.parents"]
+            for parent in parent_list:
+                yield parent,id
+
+    def create_child_link(self):
+        for parent_id,child_id in self.parent_search():
+            query="MATCH(p) WHERE '" + parent_id + "' IN p.tagsid" + " match(c) WHERE c.id='" + child_id + "' CREATE (c)-[:is_child_of]->(p)"
+        self.session.run(query)
+
+    def __call__(self):
```
alexgarel, line 167: Instead of having filename as an attribute to the class, I would propose you to have it as an argument to this method, and pass it to txt2nodes.

parser/Parser/test_Parser.py
```diff
@@ -0,0 +1,67 @@
...
+    assert text == "numero-1-n-1-des-des-francais"
+    text="Randôm Languäge wìth àccénts"
+    text=x.normalizing(text,"fr")
+    assert text == "random-language-with-accent"
+
+def test_harvest():
+    x=Parser("test")
+    x.file_name()
+    data=x.harvest()
+    test_data=["# test taxonomy",
```
alexgarel, line 42: The test is the file, here it's more common to name it `expected_data` or `expected_entries`

parser/Parser/test_Parser.py
```diff
@@ -0,0 +1,67 @@
...
+    text=x.normalizing(text,"fr")
+    assert text == "numero-1-n-1-des-des-francais"
+    text="Randôm Languäge wìth àccénts"
+    text=x.normalizing(text,"fr")
+    assert text == "random-language-with-accent"
+
+def test_harvest():
+    x=Parser("test")
+    x.file_name()
+    data=x.harvest()
```
alexgarel, line 41: I propose to name it `harvested_entries`. Data is too generic, and we get lost reading the test.

parser/Parser/test_Parser.py
alexgarel, line 42: maybe you can put it in a test.json file that you parse to get test_data (`json.load`) ? It will avoid cluttering the test, and we immediatly see it goes along with test.txt.

alexgarel (commented):
You did good progress, kudos.

I commented a lot, hoping it makes you learn.

- [ ] Also please rename parser/Parser/Parser.py to parser/parser/parser.py (no uppercase in file or directory names)

parser/Parser/Parser.py
```diff
@@ -0,0 +1,174 @@
...
+        with open(self.filename,"r",encoding="utf8") as file:
+            for line in file:
+                if counter < start : counter+=1
+                else :
+                    # sanitizing
+                    #remove any space characters at end of line
+                    line = line.rstrip()
+                    #replace ’ (typographique quote) to simple quote '
+                    line = line.replace("’", "'")
+                    #replace commas that have no space around by a lower comma character and do the same for escaped comma (preceded by a \) (to distinguish them from commas acting as tags separators)
```
aadarsh-ram, line 28: In the case of longer comments, try to reduce the length or do use the triple quote notation to make it readable.
```suggestion
                    """ replace commas that have no space around by a lower comma character and do the same for escaped comma (preceded by a \) (to distinguish them from commas acting as tags separators) """
```

parser/Parser/Parser.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def reinitialize_data(self):
+        data = { "id":'',
+        "comment":[],
+        "parent_tag":[],
+        "tags":{},
+        "tagsid":[],
+        "properties":[],
+        "value":'',
+        }
```
aadarsh-ram, line 82: Do include some indentation and spaces while declaring dictionaries, since it makes the dictionary more readable :) 
```suggestion
        data = { 
                "id" : '',
                "comment" : [],
                "parent_tag" : [],
                "tags" : {},
                "tagsid" : [],
                "properties" : [],
                "value" : '',
        }
```

aadarsh-ram (commented):
Hi! @alexgarel has already mentioned many comments regarding your code. I also have some comments and suggestions regarding your code.

parser/Parser/Parser.py
```diff
@@ -0,0 +1,174 @@
...
+                property_name, property_value = self.normalizing(line[:line.index(":")],"en"),add_line(line[line.index(":")+1:])
+                property={"id":len(data['properties'])+1,"name":property_name,"value":property_value}
+                data['properties'].append(str(property))
+
+    def txt2nodes(self):
+        """Adding nodes to database"""
+        harvested_data = self.harvest()
+        self.create_headernode(next(harvested_data))
+        for entry in harvested_data:
+            self.create_node(entry)
```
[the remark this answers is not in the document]
BryanH01: Can I ? I read one line 2 times

parser/Parser/test_Parser.py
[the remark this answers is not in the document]
BryanH01: Yes, i added "tagsid" to make it easier to find parents

parser/Parser/Parser.py
```diff
@@ -0,0 +1,174 @@
...
+            if line == '':
+                if data['id']:
+                    yield data
+                    data = self.reinitialize_data()
+                    self.block_index+=1
+            elif line[0]=="#":
+                index_comm+=1
+                data['comment'].append(self.normalizing(line,"en"))
+            elif 'stopword' in line:
+                data['id'] = "stopwords:"+str(index_stopwords)
```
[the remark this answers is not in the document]
aadarsh-ram: I think @alexgarel mentions to raise an exception in Python. I think this will help you understand about exceptions: https://docs.python.org/3/library/exceptions.html

parser/parser/test_taxonomy_parser.py
```diff
@@ -0,0 +1,86 @@
+import pytest
```
alexgarel, line 1: this file should be in parser/tests.

parser/parser/taxonomy_parser.py
```diff
@@ -0,0 +1,207 @@
...
+        self.driver = GraphDatabase.driver(uri)
+        self.session = self.driver.session() #Doesn't create error even if there is no active database
+
+    def normalized_filename(self,filename):
+        """ add the .txt extension if it is missing in the filename """
+        return filename + ( '.txt' if (len(filename)<4 or filename[-4:]!=".txt") else '')
+    
+    def file_iter(self,filename,start=0):
+        """generator to get the file line by line"""
+        counter=0
```
alexgarel, line 17: this is no use any more :-)
```suggestion
```

parser/parser/taxonomy_parser.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def new_node_data(self):
+        """ To create an empty dictionary that will be used to create node """
+        data = {
+                "id" : '',
+                "comment" : [],
+                "parent_tag" : [],
+                "tags" : {},
+                "tagsid" : [],
+                "properties" : [],
```
alexgarel, line 91: we will have to take into account https://github.com/openfoodfacts/taxonomy-editor/issues/20

parser/parser/taxonomy_parser.py
```diff
@@ -0,0 +1,207 @@
...
+                'src_position' : 0,
+        }
+        return data
+
+    def set_data_id(self,data,id,line_number):
+        if not data['id'] :
+            data['id']=id
+        else :
+            message = "Trying to overwrite an entry due to a missing new line at line " + str(line_number)
+            raise Warning(message)
```
alexgarel, line 102: raise ValueError

but I think it would be better to add your own Exception class in a "exception.py" file.

parser/parser/taxonomy_parser.py
```diff
@@ -0,0 +1,207 @@
+from neo4j import GraphDatabase
```
alexgarel, line 1: you don't need to call this file taxonomy_parser.py, parser.py is enough, the problem you had is because you don't have a __init__.py file in you parser directory.

If you add it, in test, you will be able to write:
`from parser import parser`

that said I think we should name the package taxonomy_parser (the directory) and the file should be parser.py

This is because you have to think you package name as the published name whchi should be quite unique.
(even openfoodfacts_taxonomy_parser ?)

parser/parser/test_taxonomy_parser.py
```diff
@@ -0,0 +1,86 @@
...
+    x=parser.Parser()
+    file=x.file_iter("test.txt")
+    counter=0
+    language_code_prefix = re.compile('[a-zA-Z][a-zA-Z]:')
+    for _,line in file:
+        counter+=1
+        assert line == '' or line[0]=="#" or ":" in line
+        if counter==27:
+            assert line == "carbon_footprint_fr_foodges_value:fr:10"
+    assert counter == 37+1
```
alexgarel, line 26: enumerate is your friend :-)

```suggestion
    language_code_prefix = re.compile('[a-zA-Z][a-zA-Z]:')
    for _,line, counter in enumerate(file):
        assert line == '' or line[0]=="#" or ":" in line
        if counter==26:
            assert line == "carbon_footprint_fr_foodges_value:fr:10"
    assert counter == 37
```

parser/parser/test_taxonomy_parser.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_normalizing():
+    x=parser.Parser()
+    text="Numéro #1, n°1 des ¾ des Français*"
+    text=x.normalizing(text,"fr")
+    assert text == "numero-1-n-1-des-des-francais"
+    text="Randôm Languäge wìth àccénts"
+    text=x.normalizing(text,"fr")
+    assert text == "random-language-with-accents"
```
alexgarel, line 35: can you test German (no accent removal) ?

parser/parser/test_taxonomy_parser.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_create_nodes():
```
alexgarel, line 37: You should put integration test in a different file. (even better create tests/unit and tests/integration and put each kind of file in its directory).

It's always better to distinguish them.

You should add a session fixture to remove all entries at startup (using a Parser instance if you want, or creating a session)

see https://docs.pytest.org/en/7.1.x/how-to/fixtures.html

parser/parser/test_taxonomy_parser.py
```diff
@@ -0,0 +1,86 @@
...
+    header = result.value()[0]
+    assert header == ['test-taxonomy']
+
+    # comment / preceding lines correctly added
+    query="MATCH (n) WHERE size(n.preceding_lines)>0 RETURN n.id"
+    result = x.session.run(query)
+    nodes = result.values()
+    number_of_nodes = len(nodes)
+    assert number_of_nodes == 1
+    assert nodes[0][0] == 'en:meat'
```
alexgarel, line 64: Also test n.preceding_lines value.

parser/parser/test_taxonomy_parser.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_create_child_link():
+    x=parser.Parser()
+    x.create_child_link() # nodes added in the precedent test
```
alexgarel, line 68: Never ever lean on test order. You do not have control over it…

Merge the two tests maybe. It's an integration test so it's fine to do a whole run.

parser/parser/test_taxonomy_parser.py
```diff
@@ -0,0 +1,86 @@
...
+    result = x.session.run(query)
+    number_of_links = result.value()[0]
+    assert number_of_links == 6
+
+    # correct number of links for each child node
+    query="MATCH (n)-[r]->() RETURN n.id,n.parents,count(r)"
+    results = x.session.run(query)
+    for result in results:
+        assert len(result['n.parents']) == result['count(r)']
+    assert number_of_links == 6
```
alexgarel, line 81: You really need to assert links are the correct one. I think you can build a set of tuples child / parent id from database and compare with what you expect.

alexgarel (commented):
Good progress, commenting as always ;-)

parser/Parser/test_Parser.py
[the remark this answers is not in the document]
alexgarel: @aadarsh-ram (sorry see this now) we have tags and tags_ids.

We already talk about this, but I write for the record: tags_ids is the normalized version of tags. It's important to match tags.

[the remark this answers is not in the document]
alexgarel: Now refer to 3f748b0074de71025703162a012ab979359259bf

openfoodfacts_taxonomy_parser/parser/exception.py
```diff
@@ -0,0 +1,8 @@
+class FileError(ValueError):
+    ''' Raised when attempting to change id before adding the related node i.e. when the .txt file is missing a new line '''
+
+    def __init__(self,line):
+        msg = 'missing new line at line {}'
+        exception_message = msg.format(line)
+        superinit = super().__init__
+        superinit(exception_message)
```
alexgarel, line 8: Seems exagerated to have two lines here (for a well known pattern)

```suggestion
        superinit = super().__init__(exception_message)
```

openfoodfacts_taxonomy_parser/parser/exception.py
```diff
@@ -0,0 +1,8 @@
+class FileError(ValueError):
```
alexgarel, line 1: I will use Exception (we may want to catch it independently of ValueError which is very common, and I don't really see an obvious link.) My previous comment might as misled you !

Also I don't find FileError really descriptive, while you are raising in a very specific case. Maybe DuplicateIDError would be better. (or something better ?)

```suggestion
class DuplicateIDError(Exception):
```

openfoodfacts_taxonomy_parser/parser/exception.py
```diff
@@ -0,0 +1,8 @@
...
+    ''' Raised when attempting to change id before adding the related node i.e. when the .txt file is missing a new line '''
+
+    def __init__(self,line):
+        msg = 'missing new line at line {}'
+        exception_message = msg.format(line)
```
alexgarel, line 6: Use f-string when possible

```suggestion
        exception_message = f"missing new line at line {line}"
```

openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -0,0 +1,256 @@
...
+            query += """
+                MATCH(c) WHERE c.id= $child_id
+                CREATE (c)-[:is_child_of]->(p)
+            """
+            print(query)
+            self.session.run(query , parent_id=parent_id , tagsid=tags_ids , child_id=child_id)
+
+    def delete_used_properties(self):
+        query = "MATCH (n) SET n.previous_block = null, n.parents = null"
+        self.session.run(query)
```
alexgarel, line 245: I'am not sure we really need to do this, but why not :-)

openfoodfacts_taxonomy_parser/tests/unit/test_parser_unit.py
```diff
@@ -0,0 +1,45 @@
+import pytest
+import re
+
+#importing parser
+import os,sys
+script_dir = os.path.dirname( '__file__' )
+mymodule_dir = os.path.join( script_dir, '..' , '..', 'parser')
+sys.path.append( mymodule_dir )
+import parser
```
alexgarel, line 9: you should not need this if you add a __init__.py to parser directory…

alexgarel: Also you renamed the wrong "parser" directory.

The one to name `openfoodfacts_taxonomy_parser` would be the one above `parser.py`

and then in you test you would do: `from openfoodfacts_taxonomy_parser import parser`

(sorry there again, my previous comment could have been misleading…)

alexgarel (commented):
Some comments (most are style)

alexgarel:
@BryanH01 didn't you forgot to commit the integration test file ? (Or is it still a work in progress ?)

BryanH01:
> @BryanH01 didn't you forgot to commit the integration test file ? (Or is it still a work in progress ?)

No i didn't commit it because i haven't updated it yet (because of the changes in parser)

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,283 @@
...
+                CREATE (n:TEXT {id: '__header__' })
+                SET n.preceding_lines= $header
+                SET n.src_position= 1
+            """
+        self.session.run(query,header=header)
+
+    def create_node(self,data):
+        """ run the query to create the node with data dictionary """
+        position_query = """
+            SET n.previous_block = $previous_block
```
aadarsh-ram, line 24: @BryanH01 @alexgarel Is this property called "previous_block" or "is_before"?

BryanH01: Oops sorry my bad, i forgot to change it

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,283 @@
...
+        for word in line[3:].split(","):
+            new_line.append( self.remove_stopwords( lc,self.normalizing(word,lc) ) )
+        return lc,new_line
+
+    def new_node_data(self):
+        """ To create an empty dictionary that will be used to create node """
+        data = {
+                "id" : '',
+                "main_language" : '',
+                "comment" : [],
```
aadarsh-ram, line 124: @BryanH01 @alexgarel Isn't this property is called "preceding_lines"? If so, please do change it everywhere.
```suggestion
                "preceding_lines" : [],
```

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,283 @@
...
+            words_to_remove = self.stopwords[lc]
+            new_words = []
+            for word in words.split("-"):
+                if word not in words_to_remove :
+                    new_words.append(word)
+            return ("-").join(new_words)
+        else : return words
+
+    def add_line(self,line):
+        """to get a normalized string but keeping the language code "lc:" and the "," (commas) separators , used to add an id to entries and to add a parent tag"""
```
aadarsh-ram, line 101: I think this docstring's too big. @BryanH01 would you be able to shorten it, or leave some spaces?

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,283 @@
...
+            new_line.append( self.remove_stopwords( lc,self.normalizing(word,lc) ) )
+        return lc,new_line
+
+    def new_node_data(self):
+        """ To create an empty dictionary that will be used to create node """
+        data = {
+                "id" : '',
+                "main_language" : '',
+                "comment" : [],
+                "parent_tag" : [],
```
aadarsh-ram, line 125: @BryanH01 I think this property is redundant with the "is_before" relationship. @alexgarel should this property be added inside Neo4J?

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,283 @@
...
+
+        #header
+        header,next_line = self.header_harvest(filename)
+        yield header
+        self.is_before = '__header__'
+
+        #the other entries
+        data = self.new_node_data()
+        for line_number,line in self.file_iter(filename,next_line):
+            if line == '': # can be the end of an block or just 2 line separators, file_iter() always end with ''
```
aadarsh-ram, line 170: @BryanH01 Do try to shorten the comments, since it's a little too long to read. As a  convention, do try to restrict shorter comments to the side and longer comments above a line.

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,283 @@
...
+                    data = self.new_node_data()
+                else : # there was more than 1 separator line or maybe a comment block before an entry
+                    data['comment'].append(line)
+            else :
+                if line[0]=="#":
+                    data['comment'].append(line)
+                    if not data['src_position'] : # to get the position of the footer if it's not empty
+                        data['src_position'] = line_number + 1
+                else :
+                    if len(data)==5 and not data['parent_tag'] : # the begining of the entry
```
aadarsh-ram, line 183:
```suggestion
                    if len(data)==5 and not data['parent_tag'] : # the beginning of the entry
```

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,283 @@
...
+                        data['parent_tag'].append(self.add_line(line[1:]))
+                    elif language_code_prefix.match(line):
+                        if not data['id'] :
+                            data['id']=self.add_line(line.split(',',1)[0])
+                            data['main_language']=data['id'][:2] # first 2 characters are language code
+                        #add tags and tagsid
+                        lang,line=line.split(':',1)
+                        tags_list=[]
+                        tagsids_list=[]
+                        # differenciate separating , from non-separating , by changing separators to " , "
```
aadarsh-ram, line 211:
```suggestion
                        # differentiate separating commas from non-separating commas by changing separators to " , "
```

parser/openfoodfacts_taxonomy_parser/test.txt
```diff
@@ -0,0 +1,37 @@
+# test taxonomy
+
+stopwords:fr: aux,au,de,le,du,la,a,et
+
+synonyms:en:passion fruit, passionfruit
+
+synonyms:fr:fruit de la passion, fruits de la passion, maracuja, passion
```
aadarsh-ram, line 7: @BryanH01 Use the [updated test taxonomy](https://github.com/openfoodfacts/taxonomy-editor/blob/main/backend/sample/test.txt) from main.

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,283 @@
...
+                # and do the same for escaped comma (preceded by a \) 
+                # (to distinguish them from commas acting as tags separators)
+                line = re.sub(r"(\d),(\d)", r"\1‚\2", line)
+                line = re.sub(r"\\,", "\\‚", line)
+                #removes parenthesis for roman numeral
+                line = re.sub(r"\(([ivx]+)\)", r"\1", line, flags=re.I)
+                yield line_number,line
+        yield line_number,"" #to end the last entry if not ended
+
+    def normalizing(self,line,lang="default"):
```
aadarsh-ram, line 74: @BryanH01 this whole function could be supplemented with some spaces to make it more readable.

aadarsh-ram (commented):
@BryanH01 I have added some style and property comments. Do go through them and make the necessary changes after discussion.

parser/openfoodfacts_taxonomy_parser/parser.py
BryanH01: Which one ? src_position ? If so, yes it's added in the node

parser/openfoodfacts_taxonomy_parser/parser.py
alexgarel: yes that's `preceding_lines`.

parser/openfoodfacts_taxonomy_parser/parser.py
alexgarel: @BryanH01 in docstring use line return :-)

Also you should pass your code through [black](https://pypi.org/project/black/) (and we should automate that with a make command). Because you are far from the pep 8 specs :-)

parser/openfoodfacts_taxonomy_parser/parser.py
alexgarel: @aadarsh-ram, to build the relationship, @BryanH01 must first set attributes, build relationships and then eventually remove them.

@aadarsh-ram and do not confuse "is_child_of" and "is_before" relation.
For example in the editor you should mainly care about "is_child_of", which is relation between entries.

The "is_before" is really just the order in the file.
Nodes have both is_before and is_child_of relations.

parser/openfoodfacts_taxonomy_parser/parser.py
alexgarel:
```suggestion
            if line == '': 
                # can be the end of an block or just 2 line separators, 
                # file_iter() always end with ''
```

parser/openfoodfacts_taxonomy_parser/parser.py
alexgarel: https://www.python.org/dev/peps/pep-0008/

parser/openfoodfacts_taxonomy_parser/parser.py
aadarsh-ram: @alexgarel I'm really sorry, I meant the "is_child_of" relationship in my comment. If the "parent_tag" property is being removed, there are no issues.

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,318 @@
...
+                    new_words.append(word)
+            return ("-").join(new_words)
+        else:
+            return words
+
+    def add_line(self, line):
+        """to get a normalized string but keeping the language code "lc:" , used for id and parent tag"""
+        lang = line[:2]
+        new_line = lang + ":"
+        new_line += self.remove_stopwords(lang, self.normalizing(line, lang))
```
aadarsh-ram, line 127: @BryanH01, I think this should be `line[3:]` since it doesn't build the `is_child_of` relationship without it.
```suggestion
        new_line += self.remove_stopwords(lang, self.normalizing(line[3:], lang))
```

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,336 @@
...
+            if not (line) or line[0] == "#":
+                data["preceding_lines"].append(line)
+                if not data[
+                    "src_position"
+                ]:  # to get the position of the footer if it's not empty
+                    data["src_position"] = line_number + 1
+            else:
+                if (
+                    len(data) == 5 and not data["parent_tag"]
+                ):  # the beginning of the entry
```
alexgarel, line 229: I find the `len(data) == 5` very obscure and fragile ! (it will break if I change `new_node_data`, while I won't be aware of it ! And I see many other way to unexpectedly break this !

For the footer you can compute it afterwards (you have the lenght of preceeding_line, so you can deduce where the foot start. Also it's not worth complexifying the code, just to get footer position right !

So I would remove the update of "src_position" above and here do:

```python
if not data["src_position"]:
    data["src_position"] = line_number
```

Even better, in `new_node_data`, you should put `src_position` to None and here, test with `is None`

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,336 @@
...
+                        tagsids_list.append(word_normalized)
+                    data["tags_" + lang] = tags_list
+                    data["tags_ids_" + lang] = tagsids_list
+                else:
+                    property_name, lc, property_value = line.split(":", 2)
+                    data["prop_" + property_name + "_" + lc] = property_value
+        data["id"] = "__footer__"
+        data["preceding_lines"].pop(0)
+        if not data["src_position"]:
+            data["src_position"] = line_number + 1  # to get position if it's empty
```
alexgarel, line 273: as stated above here you can do:

```suggestion
        data["src_position"] = line_number - len(data["preceding_line"])
```

parser/openfoodfacts_taxonomy_parser/test.txt
```diff
@@ -0,0 +1,37 @@
+# test taxonomy
```
alexgarel, line 1: why do we have this file here ?

parser/tests/integration/test_parser_integration.py
```diff
@@ -0,0 +1,95 @@
...
+TEST_TAXONOMY_TXT = str(pathlib.Path(__file__).parent.parent / "data" / "test")
+
+
+@pytest.fixture
+def new_session():
+    x = parser.Parser()
+    # delete all the nodes and relations in the database
+    query="MATCH (n) DETACH DELETE n"
+    x.session.run(query)
+    return x
```
alexgarel, line 15: it's strange to name it new_session and that it returns a Parser instance !

I would have written:

```suggestion
@pytest.fixture(auto_use=True)
def test_setup():
    # delete all the nodes and relations in the database
    query="MATCH (n) DETACH DELETE n"
    parser.Parser().session.run(query)
```
With autouse, you don't need to add it to tests.

It's ok to create a second Parser in your test, this is not that high a cost and it's more explicit.

parser/tests/integration/test_parser_integration.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def test_calling(new_session):
+    x=new_session
```
alexgarel, line 19: If you change your feature to `auto_use=True`, it would be 
```suggestion
def test_calling():
    x = parser.Parser()
```

alexgarel: but I would also change x to test_parser or taxonomy_parser

alexgarel: and you might have a session variable `session = test_parser.session` to avoid long line thereafter

parser/tests/integration/test_parser_integration.py
```diff
@@ -0,0 +1,95 @@
...
+    assert header == ['# test taxonomy']
+
+    # comment / preceding lines correctly added
+    query="MATCH (n:ENTRY) WHERE size(n.preceding_lines)>0 RETURN n.id,n.preceding_lines"
+    result = x.session.run(query)
+    nodes = result.values()
+    number_of_nodes = len(nodes)
+    assert number_of_nodes == 1
+    assert nodes[0][0] == 'en:meat'
+    assert nodes[0][1] == ['# meat','']
```
alexgarel, line 43: You should test at least three entry nodes and assert more than their id.

Also check one synonym and one stopwords.

parser/tests/integration/test_parser_integration.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+    #Child link test
```
alexgarel, line 47: :+1:  links part is well tested.

alexgarel (changes requested):
We are almost there, just add some more asserts in integration test, and small changes to simplify the code, if possible.

parser/openfoodfacts_taxonomy_parser/test.txt
aadarsh-ram: I also had the same doubt since I wasn't able to run this module while inside the package. This "test.txt" file should be outside `openfoodfacts_taxonomy_parser` I think.

parser/openfoodfacts_taxonomy_parser/requirements.txt
```diff
@@ -0,0 +1,3 @@
+neo4j=='4.4.5'
+re=='2.2.1'
+Unidecode=='1.3.4'
```
aadarsh-ram, line 3:
```suggestion
neo4j==4.4.5
pytz==2022.1
Unidecode==1.3.4
```

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -0,0 +1,336 @@
...
+    def remove_separating_line(self, data):
+        if data["preceding_lines"]:
+            if "synonyms" in data["id"]:
+                if "stopwords" in self.is_before:
+                    data["preceding_lines"].pop(0)
+            elif "stopwords" in data["id"]:
+                if "synonyms" in self.is_before:
+                    data["preceding_lines"].pop(0)
+            else:
+                data["preceding_lines"].pop(0)
```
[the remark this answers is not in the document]
BryanH01: Oh, this is because stopwords and synonyms often have no separating line between them, so I added the blank line in preceding_lines (when there is one) and I don't want it to be removed.
I'll add some comments

parser/tests/requirements-test.txt
```diff
@@ -0,0 +1,4 @@
+neo4j=='4.4.5'
+pytest=='7.1.2'
+re=='2.2.1'
+Unidecode=='1.3.4'
```
aadarsh-ram, line 4:
```suggestion
attrs==22.1.0
iniconfig==1.1.1
neo4j==4.4.5
packaging==21.3
pluggy==1.0.0
py==1.11.0
pyparsing==3.0.9
pytest==7.1.2
pytz==2022.1
tomli==2.0.1
Unidecode==1.3.4
```

aadarsh-ram (commented):
@BryanH01 I think your `requirements.txt` has been created wrongly, as it gave me errors while installation. In the future, do create a virtualenv, install all related packages and use the command `pip freeze > requirements.txt` to create the txt file.

parser/openfoodfacts_taxonomy_parser/test.txt
BryanH01: It's to test the parser myself when I'm writing it

alexgarel: ok, but maybe do not commit that ;-)

BryanH01: Do I alse remove the end (if __name__== ...)  ?

parser/tests/requirements-test.txt
BryanH01: Oh thank you ! I didn't know how to do it

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -175,22 +176,37 @@ def header_harvest(self, filename):
...
+            elif data["id"].startswith("stopwords"):
+                # it's a stopwords block,
+                # if the previous block is a synonyms block,
+                # there is at least one separating line
                 if "synonyms" in self.is_before:
                     data["preceding_lines"].pop(0)
+
             else:
+                # it's an entry block, there is always a separating line
                 data["preceding_lines"].pop(0)
```
alexgarel, line 210: I think your assumption here are too much based upon what you have seen so far, without any guarantee. There is no guarantee that we have new lines if we change type and so on. Remember that taxonomies are formatted by humans.
You need at list to check that `data["preceding_lines"][0]` is empty.

So I really think my proposal is the good one.
see: https://github.com/openfoodfacts/taxonomy-editor/pull/18#discussion_r932470399
If in rare cases it adds or remove a blank line, this is ok, I mean we do not lose any important information.

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -199,7 +215,9 @@ def harvest(self, filename):
         index_stopwords = 0
         index_synonyms = 0
         language_code_prefix = re.compile("[a-zA-Z][a-zA-Z][a-zA-Z]?:")
-        self.stopwords = dict()
+        self.stopwords = (
+            dict()
+        )  # it will contain a list of stopwords with their language code as key
```
alexgarel, line 220: why not simply put the comment before the line ? That's the way we do normally !

Also better use the `{}` to init a dict (not a big deal though)

```suggestion
        # stopwords will contain a list of stopwords with their language code as key
        self.stopwords = {}
```

parser/tests/integration/test_parser_integration.py
```diff
@@ -5,49 +5,98 @@
...
+        "tags_fr" : ["aux", "au", "de", "le", "du", "la", "a", "et"],
+        "preceding_lines" : []
+    }
+    for result in results:
+        node = result.value()
+        for key in expected_stopwords:
+            assert node[key] == expected_stopwords[key]
+
+
+    # entries correctly added
```
alexgarel, line 69:
```suggestion
     # entries correctly added 
     # check for two of them
```

alexgarel (changes requested):
@BryanH01 cool, thank you.

Just one thing I don't agree with. Let's discuss it in slack if needed.

parser/openfoodfacts_taxonomy_parser/parser.py
alexgarel: as you want @BryanH01

alexgarel (approved):
@BryanH01 let's merge :tada: 

I let you do the "squash and merge" when you are ready.

aadarsh-ram:
Kudos @BryanH01! :)

PR #19, opened 2022-07-16
fix: Changed JSON according to spec

aadarsh-ram:
### What
- Changed JSON schema and sample JSON according to spec

### DIFF
--- a/backend/sample/schema.json
+++ b/backend/sample/schema.json
@@ -3,33 +3,22 @@
         {
             "id" : "__header__",
             "value" : [
-                {
-                    "id" : "stopwords:IDX",
-                    "value" : {"type" : "string"},
-                    "block_index" : {"type" : "integer"},
-                    "block_subindex" : {"type" : "integer"}, 
-                    "preceding_lines" : {"type" : "integer"}
-                },
-                {
-                    "id" : "synonyms:IDX",
-                    "value" : {"type" : "string"},
-                    "block_index" : {"type" : "integer"},
-                    "block_subindex" : {"type" : "integer"}, 
-                    "preceding_lines" : {"type" : "integer"}
-                }
+                {"type" : "string"}
             ]
         },
         {
-            "id" : "__comments__",
-            "value" : [
-                {
-                    "id" : "comment",
-                    "value" : {"type" : "string"},
-                    "block_index" : {"type" : "integer"},
-                    "block_subindex" : {"type" : "integer"}, 
-                    "preceding_lines" : {"type" : "integer"}
-                }
-            ]
+            "id" : "stopwords:IDX",
+            "value" : {"type" : "string"},
+            "block_index" : {"type" : "integer"},
+            "block_subindex" : {"type" : "integer"}, 
+            "preceding_lines" : {"type" : "string"}
+        },
+        {
+            "id" : "synonyms:IDX",
+            "value" : {"type" : "string"},
+            "block_index" : {"type" : "integer"},
+            "block_subindex" : {"type" : "integer"}, 
+            "preceding_lines" : {"type" : "string"}
         },
         {
             "id" : "canonical_tagid",
@@ -44,7 +33,7 @@
             ],
             "block_index" : {"type" : "integer"},
             "block_subindex" : {"type" : "integer"},
-            "preceding_lines" : {"type" : "integer"}
+            "preceding_lines" : {"type" : "string"}
         }
     ]
 }
--- a/backend/sample/test.json
+++ b/backend/sample/test.json
@@ -3,83 +3,65 @@
         {
             "id" : "__header__",
             "value" : [
-                {
-                    "id" : "stopwords:0",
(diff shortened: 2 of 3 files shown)


backend/sample/schema.json
```diff
@@ -3,33 +3,22 @@
...
-                    "block_index" : {"type" : "integer"},
-                    "block_subindex" : {"type" : "integer"}, 
-                    "preceding_lines" : {"type" : "integer"}
-                }
-            ]
+            "id" : "stopwords:IDX",
+            "value" : {"type" : "string"},
+            "block_index" : {"type" : "integer"},
+            "block_subindex" : {"type" : "integer"}, 
+            "preceding_lines" : {"type" : "string"}
```
alexgarel, line 14: Maybe it's a list of line ? I think it would be more appropriate (although this is not central).

backend/sample/schema.json
```diff
@@ -3,33 +3,22 @@
...
+        },
+        {
+            "id" : "synonyms:IDX",
```
alexgarel, line 17: Same as stopwords, IDX is a variable.

backend/sample/schema.json
```diff
@@ -3,33 +3,22 @@
...
+            "value" : {"type" : "string"},
```
alexgarel, line 18: Same as stopwords.

backend/sample/schema.json
```diff
@@ -3,33 +3,22 @@
...
+            "block_index" : {"type" : "integer"},
+            "block_subindex" : {"type" : "integer"}, 
+            "preceding_lines" : {"type" : "string"}
         },
         {
```
alexgarel, line 23: here also we have to decompose tags and tagsid into single elements. This is important because we really want to request on this.

alexgarel (commented):
Some comments ! 

I think it would be even more important to have a file that we could import in neo4j. I don't know if you took some time to look at it already ?

backend/sample/schema.json
```diff
@@ -3,33 +3,22 @@
...
-            "value" : [
-                {
-                    "id" : "comment",
-                    "value" : {"type" : "string"},
-                    "block_index" : {"type" : "integer"},
-                    "block_subindex" : {"type" : "integer"}, 
-                    "preceding_lines" : {"type" : "integer"}
-                }
-            ]
+            "id" : "stopwords:IDX",
```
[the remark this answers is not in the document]
aadarsh-ram: Would this representation be appropriate for the "id"?
```suggestion
            "id" : {
                "type" : "string",
                "const": "stopwords:IDX"
            }
```

backend/sample/schema.json
```diff
@@ -3,33 +3,22 @@
...
+            "value" : {"type" : "string"},
```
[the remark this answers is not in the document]
aadarsh-ram: > It's a list of tag and tagid. Something like `[{"tag": "en:The", "tagid": "en:the"}, {"tag: "en:Some", "tagid": "en:some"}]`.

I think this would be a better representation, since each tag and tagid would be in one dictionary and easy to query.

backend/sample/schema.json
aadarsh-ram: Yep, I'll add that.

backend/sample/schema.json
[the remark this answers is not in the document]
alexgarel: The point is that it's not a const, but, I tihnk it's ok for now, as this is mainly to document.
so it's just `

```suggestion
            "id" : {
                "type" : "string",
                "description": "stopwords:IDX"
            },
```

backend/sample/schema.json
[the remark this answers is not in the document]
alexgarel: For stopwords we only have one language per entry so maybe something like:
```
...
"lang": "en", 
"tag": ["The", "Some"], 
"tagid": ["the", "some"]
...
```
is the right way.

backend/sample/test.json
```diff
@@ -59,7 +77,10 @@
                 "en" : ["Passion fruit yogurts"],
                 "fr" : ["yaourts au fruit de la passion"]
             },
-            "tagsid" : ["en:Passion fruit yogurts", "fr:yaourts au fruit de la passion"],
+            "tagsid" : {
+                "en" : ["Passion fruit yogurts"],
+                "fr" : ["yaourts au fruit de la passion"]
+            },
```
alexgarel, line 83: In tagsid you have to normalize spaces and remove stopwords for the language.

I let you do it in the other entries ;-)

```suggestion
            "tagsid" : {
                "en" : ["passion-fruit-yogurts"],
                "fr" : ["yaourts-fruit-passion"]
            },

```

aadarsh-ram: Oh okay, I'll check which entries have stopwords and normalize them.

alexgarel (commented):
@aadarsh-ram let's merge.

On my side I will do a PR with:
- a neo4j importable file
- spec update
- eventual update to those files

alexgarel (approved):

PR #26, opened 2022-07-25
feat: New paths for backend API

### LINKED ISSUE #24: Add GET methods in backend API
Create GET paths in the backend API for retrieving info on all node types in the taxonomy.

### LINKED ISSUE #25: Add POST methods for backend API
Create POST methods in the backend API for updating info on all node types in the taxonomy.

aadarsh-ram:
## What
- Creates new paths for retrieval and updation of nodes in a taxonomy
- List of GET paths created:

| Paths           | Description      |
|:--------------|:-------------------------------------------------|
| "/nodes" | Retrieves all node information in taxonomy |
| "/entry/{entry}" | Retrieves information on entry with id = {entry} |
| "/entry" | Retrieves all nodes with label "ENTRY" |
| "/synonym/{synonym}" | Retrieves information on synonym with id = {synonym} |
| "/synonym" | Retrieves all nodes with label "SYNONYMS" |
| "/stopword/{stopword}" | Retrieves information on stopword with id = {stopword} |
| "/stopword" | Retrieves all nodes with label "STOPWORDS" |
| "/header" | Retrieves information on header of taxonomy |
| "/footer" | Retrieves information on footer of taxonomy |
    
 - List of POST paths created:
 
| Paths           | Description      |
|:--------------|:-------------------------------------------------|
| "/edit/entry/{entry}" | Editing an entry in taxonomy with id = {entry} |
| "/edit/synonym/{synonym}" | Editing a synonym with id = {synonym} |
| "/edit/stopword/{stopword}" | Editing a stopword with id = {stopword} |
| "/edit/header" | Editing header of taxonomy |
| "/edit/footer" | Editing footer of taxonomy |

## Part of
- Fixes #24 
- Fixes #25


backend/editor/api.py
```diff
@@ -25,6 +27,26 @@
...
 
+@app.on_event("startup")
+async def startup():
+    """
+    Initialize Neo4J database
+    """
+    global driver, session
+    uri = "bolt://localhost:7687"
+    driver = GraphDatabase.driver(uri)
+    session = driver.session()
```
alexgarel, line 38: well done. 

Could we take the uri from environment and put this env reading in a settings.py file (just to group them, also because it's easier to override them for test)

That is you will have setting.py with

```
```

and in api.py:

```
import settings

uri = settings.uri
```

(it's intentional not to write `from settings import uri` but `import settings` to ease monkeypatching it for tests, if needed)

backend/editor/api.py
```diff
@@ -35,4 +57,214 @@ async def pong(response: Response):
...
+@app.get("/nodesfull")
+async def findAllNodes(response: Response):
+    """
+    Get all nodes within taxonomy
+    """
+    query = """
+        MATCH (n) RETURN n
+    """
+    result = session.run(query)
+    allNodes = [record for record in result]
```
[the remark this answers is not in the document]
alexgarel: I let you apply it everywhere else !

backend/editor/api.py
```diff
@@ -35,4 +57,214 @@ async def pong(response: Response):
...
+    """
+    Get entry corresponding to id within taxonomy
+    """
+    query = """
+        MATCH (n:ENTRY) WHERE n.id = $id 
+        RETURN n
+    """
+    result = session.run(query, {"id": entry})
+    oneEntry = [record for record in result]
+    return oneEntry
```
alexgarel, line 85: you just want one entry. So it's better not to return the list.

I would write:
```suggestion
    oneEntry = next(iter(result))
    return oneEntry
```

but you also should handle the case where there is no such entry and return a 404.
```

File: backend/editor/api.py [hunk 4]
```diff
@@ -35,4 +57,214 @@ async def pong(response: Response):
...
+@app.get("/entry/{entry}")
+async def findOneEntry(response: Response, entry: str):
+    """
+    Get entry corresponding to id within taxonomy
+    """
+    query = """
+        MATCH (n:ENTRY) WHERE n.id = $id 
+        RETURN n
+    """
+    result = session.run(query, {"id": entry})
```
alexgarel, line 83: It would be better to separate database access from http processing.

So it would be better to have a entries.py where you put:
- the session initialization / shutdown (but events handlers should be here, they just call the right function)
- all the requests as simple functions (eg, here: `get_entry(id)`)

It's always better to separate that kind of things:
- you can factor out something (here for example I imagine only one function get_node(id, type), and in findOneEntry, we can call `get_node(id, "ENTRY")° (BTW "ENTRY" could be in an enumeration)
- for testability (you can test db function on their own)
- as you code will grow you will have a better [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) (and this is a very important design pattern in my experience)

backend/editor/api.py
```diff
@@ -35,4 +57,214 @@ async def pong(response: Response):
...
+    """
+    incomingData = await request.json()
+    result = None
+    for key in incomingData.keys():
+        query = f"""
+            MATCH (n:ENTRY) WHERE n.id = $id
+            SET n.{key} = $value
+            RETURN n
+        """
+        result = session.run(query, {"id": entry, "value": incomingData[key]})
```
alexgarel, line 193: You should do it all in one query, this will be much more efficient.

You just have to build the "SET" lines and concatenate them. For each line instead of $value you can use the key name `f"SET .... ${key}"` and pass the whole dict to "run", adding the "id".

backend/editor/api.py
```diff
@@ -35,4 +57,214 @@ async def pong(response: Response):
...
+        MATCH (n:TEXT) WHERE n.id = "__footer__"
+        RETURN n
+    """
+    result = session.run(query)
+    footer = [record for record in result]
+    return footer
+
+# Put methods 
+
+@app.put("/edit/entry/{entry}")
```
alexgarel, line 178: You did a "PUT" while you method reminds more of a "POST" in the sens that it permits partial updates (normally POST enables partial updates, but not PUT). see https://en.wikipedia.org/wiki/Representational_state_transfer#Semantics_of_HTTP_methods

backend/editor/api.py
alexgarel, line 193: Here also, if we have a entries.py or neo4j.py file, we could factorize the code of all those methods in a `update_entry(id, type)` method.

backend/editor/api.py
```diff
@@ -35,4 +57,214 @@ async def pong(response: Response):
...
+            MATCH (n:STOPWORDS) WHERE n.id = $id
+            SET n.{key} = $value
+            RETURN n
+        """
+        result = session.run(query, {"id": stopword, "value": incomingData[key]})
+    updatedStopword = [record for record in result]
+    return updatedStopword
+
+class Header(BaseModel):
+    preceding_lines: List
```
alexgarel, line 236: put models into their own file.

alexgarel: Also I would do a subclass of BaseModel with preceding_lines which would be a parent of Footer and Header.

alexgarel (changes requested):
Cool @aadarsh-ram , you are on the right path. 

Although I would have concentrated first on "entry" not the rest, because here you run the risk of having to fix lots of same errors on each API, while you could learn first on entries, and then do the rest (it's just an advice).

There are things to separate and factorize.

backend/editor/api.py
alexgarel: > you also should handle the case where there is no such entry and return a 404

Forgot to write that with proposed code it can be handled with a try / except.

But as you may also check you have only one item, you may also convert to list and the check list `len` and act accordingly.

backend/editor/api.py
alexgarel: see how I did this in sample/load.py

backend/editor/api.py
aadarsh-ram: Would this check be required for the __header__ and __footer__ nodes too? Since in a taxonomy, it is understood that they are always present.

alexgarel: Maybe not.

backend/editor/api.py
aadarsh-ram: @alexgarel I'm not able to figure out where exactly you've written the "concatenation" step in load.py. Would you be able to point that out to me?

aadarsh-ram: I built the "SET" lines by concatenation and used it as a query.

backend/editor/api.py
aadarsh-ram: Yep, I wasn't sure of the method for partial updates. I'll change it.

backend/editor/entries.py
```diff
@@ -0,0 +1,60 @@
...
+
+def update_nodes(label, entry, incomingData):
+    query = f"""MATCH (n:{label}) WHERE n.id = $id"""
+
+    for (key,value) in incomingData.items():
+        query += f"""\nSET n.{key} = {value}\n"""
+
+    query += f"""RETURN n"""
+
+    result = session.run(query, {"id": entry})
```
alexgarel, line 49: It's a bit silly, because we already inject keys, but non the less, I will use parameters to inject values.

To avoid [injection](https://developer.mozilla.org/en-US/docs/Glossary/SQL_Injection) problems, we may want to check keys only use word.

Last: when you build a query like this, the best approach is to collect lines in a list and join them at the end (faster, cleaner)

So I propose:

```suggestion
    # sanity check keys
    for key in incomingData.keys():
        if not re.match(r"^\w+$", key) or key == "id":
            raise ValueError("Invalid key: %s", key)
    # build query
    query = [f"""MATCH (n:{label}) WHERE n.id = $id"""]
    for key in incomingData.keys():
        query.append(f"""\nSET n.{key} = ${key}\n""")
    query.append(f"""RETURN n""")
    params = dict(incomingData, id=entry)
    result = session.run(" ".join(query), params)
```

backend/editor/entries.py
```diff
@@ -0,0 +1,60 @@
...
+    return result
+
+def update_marginals(id, incomingData):
+    convertedData = incomingData.dict()
+    query = f"""
+        MATCH (n:TEXT) WHERE n.id = $id
+        SET n.preceding_lines = {str(convertedData['preceding_lines'])}
```
alexgarel, line 56: There again there is a risk of injection.

backend/editor/entries.py
```diff
@@ -0,0 +1,60 @@
...
+        RETURN n
+    """
+    result = session.run(query, {"id": id})
+    return result
```
alexgarel, line 60: In this case also I will simply reuse update_nodes, it's the simplest.

alexgarel (changes requested):
Really good @aadarsh-ram 

I still request quite a lot of change so that we have a perfect code :-)

backend/editor/dependencies.py
```diff
@@ -0,0 +1,16 @@
...
+# FastAPI
+from fastapi import FastAPI, status, Response, Request, HTTPException
+from fastapi.responses import JSONResponse
+from fastapi.middleware.cors import CORSMiddleware
+
+# Pydantic: Define data schema
+from pydantic import BaseModel
+
+# Data model imports
+from models import Header, Footer
```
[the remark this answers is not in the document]
aadarsh-ram: Yeah, it was based on the structure followed in folksonomy_api. It seemed clean to use a separate dependencies file.

backend/editor/entries.py
```diff
@@ -0,0 +1,60 @@
+from dependencies import *
+
+def initialize_db():
```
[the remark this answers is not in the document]
aadarsh-ram: Since, a doc string was already present in the api.py file, I thought it was self-explanatory. But, I'll add a description over here too.

backend/editor/api.py
```diff
@@ -1,12 +1,5 @@
-from datetime import datetime
-from typing import List, Optional
-from neo4j import GraphDatabase
-
-# FastAPI
-from fastapi import FastAPI, status, Response, Request
-from fastapi.responses import JSONResponse
-from fastapi.middleware.cors import CORSMiddleware
-from pydantic import BaseModel
+from dependencies import *
```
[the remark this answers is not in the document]
aadarsh-ram: So, should I bring all required imports to their respective files and remove the dependencies.py file?

backend/editor/dependencies.py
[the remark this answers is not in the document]
alexgarel: Ooooh, ok… but I don't see it as a good design pattern.

(We will introduce flake8 soon).

backend/editor/api.py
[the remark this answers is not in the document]
alexgarel: > So, should I bring all required imports to their respective files and remove the dependencies.py file?

yes please.

backend/editor/entries.py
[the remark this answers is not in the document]
alexgarel: yep, doc string to every function is a good habit ! (did I repeat myself ? :stuck_out_tongue_closed_eyes: )

backend/editor/api.py
```diff
@@ -64,113 +53,92 @@ async def findAllNodes(response: Response):
...
-        MATCH (n:ENTRY) WHERE n.id = $id 
-        RETURN n
-    """
-    result = session.run(query, {"id": entry})
-    oneEntry = [record for record in result]
+    result = get_nodes("ENTRY", entry)
+    oneEntry = list(result)
+
+    if len(oneEntry) == 0:
+        raise HTTPException(status_code=404, detail="Entry not found")
```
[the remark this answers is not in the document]
aadarsh-ram: I think writing a `check_one()` function should be fine. Do you think we can raise a 404 with a different message like "Multiple entries found"?

backend/editor/api.py
```diff
@@ -1,11 +1,23 @@
...
+from fastapi import FastAPI, status, Response, Request, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
 
+# Data model imports
+from models import Header, Footer
+
+# DB helper imports
+from entries import initialize_db, shutdown_db
+from entries import get_all_nodes, get_nodes
+from entries import update_nodes
```
alexgarel, line 18: sorry I didn't spot it before but:

* you need a __init__.py file in editor folder (this will make the editor folder a package)
* use "relative" import here:

```suggestion
from .models import Header, Footer

# DB helper imports
from .entries import initialize_db, shutdown_db
from .entries import get_all_nodes, get_nodes
from .entries import update_nodes
```

It's important to avoid collision with eventual system level packages.

alexgarel (commented):
Just one single comment :-)
You understood previous comments very well.

backend/editor/api.py
aadarsh-ram: A init.py file is already present in the repo. For testing, it was easier to not have relative imports.

backend/editor/api.py
```diff
@@ -35,4 +74,152 @@ async def pong(response: Response):
...
+    """
+    Get __footer__ within taxonomy
+    """
+    result = get_nodes("TEXT", "__footer__")
+    footer = list(result)
+    return footer[0]
+
+# Post methods
+
+@app.post("/edit/entry/{entry}")
```
alexgarel, line 171: Sorry for I didn't check it.

Usually on APIs the best is to just have 

```suggestion
@app.post("/entry/{entry}")
```

It's a common design pattern, that with a GET you have the entry, with a POST you modify it.

I think it's worth it, if you agree @aadarsh-ram

aadarsh-ram: Yep, I agree. I'll make the changes

alexgarel (approved):
Great work @aadarsh-ram,

Let's merge :-)

Issue #28, opened 2022-07-26
Add main_language as a property to the entries in neo4j


alexgarel:
Currently we indirectly have this information through the language code in the id.

But we want to be able to change main language (that is language giving the id) without modifying the id (yet) to make things simpler. (we shall change our mind on this later).

Add a "main_language" property to entries.

- [ ] change documentation and sample
- [ ] change parser

alexgarel:
@BryanH01 I let you close this ticket when you have introduced the property in the parser.

BryanH01:
I added it

PR #29, opened 2022-07-26
docs: introduce main_language property on entries

alexgarel:
Documentation + sample part for
- #28


BryanH01 (approved):

aadarsh-ram:
@alexgarel @stephanegigandet Do you think we might need a main_language property for stopwords and synonyms too?

alexgarel:
> Do you think we might need a main_language or a "language" property for stopwords and synonyms too?

@aadarsh-ram I think for now we don't need it. We will do very few operations on those entries. First the editor should concentrate on entries.

PR #31, opened 2022-07-29
feat: Frontend in React

### LINKED ISSUE #34: React frontend for updation and retrieval of entries
As disscussed with @alexgarel @stephanegigandet, a first iteration of the React frontend for editing and retrieval of entries in a taxonomy will be implemented.

aadarsh-ram:
WARNING: this PR is being dispatched in more than one PR.

### What
- Setup basic React project
- Build pages for retrieval and updation of entries

### Related issue(s)
- Fixes #34


aadarsh-ram:
@alexgarel I'll be working on this PR for the frontend.

taxonomy-editor-frontend/src/App.js
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <Router>
+      <div className="App">
+        <Routes>
+          <Route path="/" element={<Home />} />
+          <Route path="/entry" element={<Entry />} />
+          <Route path="/entry/:id" element={<EditEntry />} />
+        </Routes>
+      </div>
+    </Router>
```
alexgarel, line 16: I don't know if it can be moved easily later on (and in this case we can let it like that for now), but I think in the long run, we will have something like: "/:branch-name/:taxonomy/entry/:id"
- branch-name would be the name of a branch in git
- taxonomy would be taxonomy name (like categories)

aadarsh-ram: @alexgarel Yep, I think it would be possible. The "branch-name" property might be a little difficult to bring in right away, due to its complexity related to auth and error handling. The "taxonomy" property can be brought in, whenever the provision of seperate databases for each taxonomy is brought in Neo4J.

taxonomy-editor-frontend/src/components/useFetch.jsx
```diff
@@ -0,0 +1,36 @@
+import { useState, useEffect } from "react";
+
+const useFetch = (url) => {
```
nobeeakon, line 3: NOT relevant for this PR, but what about just using a library for all the fetching and stuff: [query](https://tanstack.com/query/v4)  does an excelent job on that and it is simple to use

taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+const EntriesList = ({ nodes, title }) => {
+    const navigate = useNavigate();
+    const handleClick = (event, id) => {
+        event.preventDefault();
+        navigate('/entry/'+id);
+    }
+    const theme = createTheme({
+        typography: {
+            fontFamily : 'Plus Jakarta Sans',
```
nobeeakon, line 21: maybe use this at the top of the tree? ... like in the App.jsx ?, that way you get a general style for all the app

taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
```diff
@@ -0,0 +1,84 @@
...
+                    List of nodes in {title} Taxonomy:
+                </Typography>
+                <div className="entry-preview">
+                    <TableContainer sx={{ml: 2}} component={Paper}>
+                        <Table sx={{ width: 500 }}>
+                            <TableHead>
+                            <TableRow>
+                                <TableCell align="center">
+                                <Typography variant="h6">
+                                    #
```
nobeeakon, line 37: is this needed?

taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
```diff
@@ -0,0 +1,84 @@
...
+                            </TableHead>
+                            <TableBody>
+                                {nodes.map((node, index) => (
+                                    <TableRow
+                                    key={node[0].id}
+                                    sx={{ '&:last-child td, &:last-child th': { border: 0 } }}
+                                    >
+                                        <TableCell align="center" component="th" scope="row">
+                                            <Typography variant="subtitle1">
+                                                {index+1}
```
nobeeakon, line 60: is this needed?

taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
```diff
@@ -0,0 +1,84 @@
...
+                                </TableCell>
+                            </TableRow>
+                            </TableHead>
+                            <TableBody>
+                                {nodes.map((node, index) => (
+                                    <TableRow
+                                    key={node[0].id}
+                                    sx={{ '&:last-child td, &:last-child th': { border: 0 } }}
+                                    >
+                                        <TableCell align="center" component="th" scope="row">
```
nobeeakon, line 58: nit: `component = 'td'` [th vs td](TH is used for table header cells while TD is used for table data cells)

taxonomy-editor-frontend/src/pages/allentries/index.jsx
```diff
@@ -0,0 +1,16 @@
+import useFetch from "../../components/useFetch";
+import EntriesList from "./EntriesList";
+
+const Entry = () => {
+    const {data: nodes, isPending, error} = useFetch('http://localhost:80/nodes');
+    // console.log(nodes, isPending, error);
+    return (
+        <div className="allentries">
+            {error && <div>{error}</div>}
```
nobeeakon, line 9: In general I'd prefer to split this... something like: 
`if (error) return ( <div className="allentries">{error} </div>`
and so on. That way even if the main content grows, those other simple cases (error, loading) are kept simple and visible

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,93 @@
...
+    const [nodeObject, setNodeObject] = useState(null);
+    useEffect(() => {
+        setNodeObject(props);
+    }, [props])
+
+    let lc = '';
+    let toBeRendered = {}
+    let languageNames = new Intl.DisplayNames(['en'], {type: 'language'});
+
+    if (nodeObject) {
```
nobeeakon, line 14: I may not be getting the whole idea , but, maybe pass this to the useEffect ? and set the state of line 5 to what you will use at the end?

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+    function changeData(key, index, obj) {
+        console.log(key, index, obj);
+        const duplicateData = {...nodeObject};
+        duplicateData[key][index] = obj;
+        setNodeObject(duplicateData);
+    }
+
+    return ( 
+        <div className="allProperties">
```
nobeeakon, line 39: better avoid using general css classNames.

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,93 @@
...
+            <Typography sx={{ml: 4, mt: 2, mb: 1, textDecoration: 'underline'}} variant='h5'>Comments</Typography>
+            { nodeObject && <TextField
+                sx={{ml: 8, mt: 1}}
+                size="small"
+                multiline
+                onChange = {event => {
+                    // console.log(event.target.value);
```
nobeeakon, line 46: please remove console.log

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,93 @@
...
+import { useEffect, useState } from "react";
+
+const ListAllOtherProperties = ({ props }) => {
+    const [nodeObject, setNodeObject] = useState(null);
+    useEffect(() => {
+        setNodeObject(props);
+    }, [props])
+
+    let lc = '';
+    let toBeRendered = {}
```
nobeeakon, line 11: if possible, please improve the naming of these two variables.

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,93 @@
...
+                </div>
+                }
+            
+            { Object.keys(toBeRendered).length > 0 && 
+                    <div className="tags">
+                        {nodeObject.id.startsWith('stopword') && 
+                            <Typography sx={{ml: 4, mt: 1, mb: 1, textDecoration: 'underline'}} variant='h5'>
+                                Stopwords
+                            </Typography>}
+                        {nodeObject.id.startsWith('synonym') &&
```
nobeeakon, line 70: nit: guess these two are excluse, what about using  the ternary?

taxonomy-editor-frontend/src/pages/editentry/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    return (
+        <ThemeProvider theme={theme}>
+            <Typography sx={{mb: 2, mt:2, ml: 2}} variant="h4">
+                You are now editing "{id}" 
+            </Typography>
+            { isEntry && <FetchRelations url={url+'parents'} title={'Parents'} /> }
+            { isEntry && <FetchRelations url={url+'children'} title={'Children'} /> }
+            { isEntry && <ListAllEntryProperties props={nodeObject} /> }
```
nobeeakon, line 38: what about having all of those wrapped in <> </> or in a <div> ?

taxonomy-editor-frontend/src/pages/editentry/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+    else if (id.startsWith('stopword')) { url += `stopword/${id}/` }
+    else { url += `entry/${id}/`; isEntry = true; }
+
+    const { data: node, error, isPending } = useFetch(url);
+    let nodeObject = null;
+    if (!isPending) {
+        nodeObject = node[0];
+    }
+
+    const theme = createTheme({
```
nobeeakon, line 25: better avoid having partial small custom themes. If possible move it to the top of the tree

taxonomy-editor-frontend/src/pages/home/index.jsx
```diff
@@ -0,0 +1,83 @@
+import Button from '@mui/material/Button';
+import CssBaseline from '@mui/material/CssBaseline';
+import Link from '@mui/material/Link';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import Container from '@mui/material/Container';
+import { createTheme, ThemeProvider } from '@mui/material/styles';
+import { useNavigate } from "react-router-dom";
+
+function Copyright(props) {
```
nobeeakon, line 10: if this component is only used here, what about not passing props and defining everything on it? passing props and descructuring this way becomes a bit difficult to follow at some point

taxonomy-editor-frontend/src/pages/home/index.jsx
```diff
@@ -0,0 +1,83 @@
...
+        Open Food Facts
+      </Link>{' '}
+      {new Date().getFullYear()}
+      {'.'}
+    </Typography>
+  );
+}
+
+const Home = () => {
+    const theme = createTheme({
```
nobeeakon, line 24: same as with other createTheme , please check the other comments on this :)

taxonomy-editor-frontend/src/App.js
```diff
@@ -0,0 +1,24 @@
...
+import ResponsiveAppBar from "./components/ResponsiveAppBar";
+import Entry from "./pages/allentries";
+import EditEntry from "./pages/editentry";
+import Home from './pages/home';
+
+function App() {
+  return (
+    <Router>
+      <div className="App">
+        <CssBaseline />
```
nobeeakon, line 12: CssBaseline is ok here, no need for the others

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
```diff
@@ -0,0 +1,69 @@
...
+        duplicateData[key] = value;
+        setChangedObj(duplicateData);
+    }
+    
+    // Helper function used for changing state of comment
+    function changeDataComment(key, value) {
+        const duplicateData = {...changedComment};
+        duplicateData[key] = value;
+        setChangedComment(duplicateData);
+        // console.log(changedComment);
```
nobeeakon, line 29: please remove console logs

taxonomy-editor-frontend/src/pages/home/index.jsx
```diff
@@ -0,0 +1,83 @@
...
+                component="img" 
+                width={128} 
+                height={128}
+                src={require('../../assets/classification.png')} 
+                alt="Classification Logo" 
+            />
+            <Typography sx={{mt: 4, mb: 4}} variant="h2">
+                Taxonomy Editor
+            </Typography>
+            <Box component="form" onSubmit={ handleSubmit } noValidate sx={{ mt: 1 }}>
```
nobeeakon, line 66: better use a Link provided by react-router for navigation

taxonomy-editor-frontend/src/pages/home/index.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+const Home = () => {
+    const theme = createTheme({
+        typography: {
+            fontFamily : 'Plus Jakarta Sans',
+        },
+    });
+    const navigate = useNavigate();
+
+    const handleSubmit = (event) => {
```
nobeeakon, line 31: please check the comment below

taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
aadarsh-ram: @nobeeakon It was just a way to keep track of the number of entries in the table. It isn't necessary, but the table seemed to be clear with it. Do you think it needs to be removed?

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
aadarsh-ram: Since setNodeObject is async, it isn't possible to traverse through it's keys inside useEffect. So, I'm checking it separately after the useEffect.

Also, I'm maintaining two separate objects (`nodeObject` and `toBeRendered`) so that after the user has done all the edits, I'll be able to update nodeObject using values from `toBeRendered` and give a POST request of nodeObject to the backend. Does this make sense?

taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
nobeeakon: yeah @aadarsh-ram, not sure on that... I didn't run the code so I'm not completly sure how this looks like. On the other side, I think this is more of a UX decision, if this info is providing the user some additional/useful information it is good to keep. But, if the index is not providing any info the user actually needs or uses I'd remove it as doing so would save some space. If what you want is to have just a counter what about adding a small legend at the bottom that says how many rows the table has?

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,92 @@
+import { Paper, Stack, TextField, Typography } from "@mui/material";
+import { useEffect, useState } from "react";
+
+const ListAllOtherProperties = ({ props }) => {
```
nobeeakon, line 4: nit: prop called `props` could be renamed to something more explicit

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
nobeeakon: yeah makes sense, thanks

taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
aadarsh-ram: Yep @nobeeakon, I think that would work. Didn't think about it that way. I'll make the necessary changes.

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,92 @@
...
+    // Storing keys and values that needs to be rendered for editing
+    let toBeRendered = {} 
+    // Used for conversion of LC to long-form
+    let languageNames = new Intl.DisplayNames(['en'], {type: 'language'}); 
+
+    if (nodeObject) {
+        Object.keys(nodeObject).forEach((key) => {
+            if (key.startsWith('tags') && 
+                !key.includes('ids')) {
+                    languageCode = key.slice(-2);
```
nobeeakon, line 21: nit: maybe add a comment here of how the key looks like? I mean, it would help to remember why this is useful and what result it will produce

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,92 @@
...
+                    toBeRendered[languageCode] = nodeObject[key]
+                }
+        })
+    }
+
+    // Helper function used for changing state of properties
+    function changeDataComment(key, value) {
```
nobeeakon, line 28: nit: if this is only used for the comment you could remove the key and just pass the value

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    function changeData(key, index, obj) {
+        const duplicateData = {...nodeObject};
+        duplicateData[key][index] = obj;
+        setNodeObject(duplicateData);
+    }
+
+    return ( 
+        <div className="all-other-properties">
```
nobeeakon, line 41: nit: avoid having general classes, it gets tricky to maintain at some point. What about using <Box sx={{}} /> or some other Material UI option instead?

aadarsh-ram: @nobeeakon Would you be able to suggest any other alternatives?

nobeeakon: @aadarsh-ram , I have updated the previous comment, sorry for that... about other alternatives, mmm, I'm not completely familiar with the lastest version of Material, but they suggest some stuff in [here](https://mui.com/material-ui/customization/how-to-customize/)

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -0,0 +1,52 @@
+import { Button } from "@mui/material";
+import { useEffect, useState } from "react";
+import { useNavigate } from "react-router-dom";
+import FetchRelations from "./FetchRelations";
+import ListAllEntryProperties from "./ListAllEntryProperties";
+import ListAllOtherProperties from "./ListAllOtherProperties";
+import Swal from 'sweetalert2'
+import withReactContent from 'sweetalert2-react-content'
```
nobeeakon, line 8: what about using [Material ui dialog](https://mui.com/material-ui/api/dialog/)? think it is better to stick with that for styling and so on, also removes the need of an additional library (more things to pass to the client)

taxonomy-editor-frontend/package.json
```diff
@@ -16,6 +16,8 @@
     "react-i18next": "^11.18.3",
     "react-router-dom": "^6.3.0",
     "react-scripts": "5.0.1",
+    "sweetalert2": "^11.4.24",
+    "sweetalert2-react-content": "^5.0.2",
```
nobeeakon, line 20: please see the comment below

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -0,0 +1,52 @@
...
+            method : 'POST',
+            headers: {"Content-Type" : "application/json"},
+            body: JSON.stringify(nodeObject)
+        }).then(() => {
+            SuccessSwal.fire({
+                title: <strong>Your edit was successful!</strong>,
+                icon: 'success'
+              }).then(() => {
+                return navigate('/entry')
+              })
```
nobeeakon, line 32: maybe add a catch(error) ? even if you do nothing with the error

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -0,0 +1,52 @@
...
+        })
+    }
+    
+    return ( 
+        <div className="node-attributes">
+            { isEntry && <FetchRelations url={url+'parents'} title={'Parents'} /> }
+            { isEntry && <FetchRelations url={url+'children'} title={'Children'} /> }
+            { isEntry && <ListAllEntryProperties nodeObject={nodeObject} setNodeObject={setNodeObject} /> }
```
nobeeakon, line 40: nit: this could be:`{isEntry? <> <Fetch ...> <Fetch...> <List ...> : <ListAllOther...>` that way it is easier to read and you get the sense that those are mutually exclusive

taxonomy-editor-frontend/src/pages/editentry/index.jsx
```diff
@@ -4,6 +4,7 @@ import FetchRelations from "./FetchRelations";
 import useFetch from "../../components/useFetch";
 import ListAllEntryProperties from "./ListAllEntryProperties";
 import ListAllOtherProperties from "./ListAllOtherProperties";
+import AccumulateAllComponents from "./AccumulateAllComponents";
 
 const EditEntry = () => {
     let url = 'http://localhost:80/'
```
nobeeakon, line 10: what about passing this to a constants.js file? or maybe even an env variable?, thus you don't need to pass it all the way down. (don't know if this is the case, but for [ create-react-app ](https://create-react-app.dev/docs/adding-custom-environment-variables/#adding-development-environment-variables-in-env) )

taxonomy-editor-frontend/src/pages/editentry/index.jsx
```diff
@@ -19,9 +20,7 @@ const EditEntry = () => {
     const { data: node, error, isPending } = useFetch(url);
     let nodeObject = null;
```
nobeeakon, line 21: should this be an state?, don't quite follow why this is a `let`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -12,16 +11,13 @@ const ListTranslations = ({ nodeObject, languageNames, setNodeObject }) => {
             }
     })
 
-    // Add main language before setting it as state variable
-    let duplicateToBeRendered = {...toBeRendered};
-    duplicateToBeRendered[nodeObject.main_language] = nodeObject["tags_"+nodeObject['main_language']];
-    const [changedObj, setChangedObj] = useState(duplicateToBeRendered);
-
     // Helper function used for changing state
     function changeData(key, index, obj) {
```
nobeeakon, line 15: is the obj really an `{}` ? from what happens below I'd think it is a string?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -12,16 +11,13 @@ const ListTranslations = ({ nodeObject, languageNames, setNodeObject }) => {
...
-        const duplicateData = {...changedObj};
+        key = 'tags_' + key;
+        const duplicateData = {...nodeObject};
         duplicateData[key][index] = obj;
-        setChangedObj(duplicateData);
+        setNodeObject(duplicateData);
+        console.log(nodeObject);
```
nobeeakon, line 20: please remove the `console.log` if not needed

taxonomy-editor-frontend/src/pages/editentry/index.jsx
aadarsh-ram: Since I have to declare `nodeObject` conditionally (based on value of `isPending`), I can't define it using a useState hook. Hence, I'm updating the variable after a conditional statement. Here, it isn't a state variable, but in `AccumulateAllComponents.js` it is.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
aadarsh-ram: Yeah, it's a string. Sorry for the bad naming of the variable.

taxonomy-editor-frontend/src/pages/editentry/index.jsx
aadarsh-ram: I can create a constants.js file, but I can't avoid passing the url since the URL for accessing each entry is dependant on it's id.

taxonomy-editor-frontend/src/pages/editentry/index.jsx
nobeeakon: yep, I mean, you could pass it directly instead of creating a `let` , for example `incomingNodeObject={node?.[0]}` . Having the `let` makes it a bit confusing to read and in this case I think it may not be necessary

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+    useEffect(() => {
+        setNodeObject(incomingNodeObject);
+    }, [incomingNodeObject])
+
+    const handleClose = () => {setOpen(false)};
+    const handleBack = () => {navigate('/entry')};
+
+    const handleSubmit = () => {
+        delete nodeObject['id']; // ID not allowed in POST
```
nobeeakon, line 21: not the best thing to modify a state directly, it may lead to some unexpected behavior. What about using something like `const {id, ...data} = nodeObject` and you pass the `data`

taxonomy-editor-frontend/src/pages/editentry/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+const EditEntry = () => {
+    let url = API_URL;
+    let isEntry = false;
+    const { id } = useParams();
+    if (id.startsWith('__header__')) { url += 'header/' }
+    else if (id.startsWith('__footer__')) { url += 'footer/' }
+    else if (id.startsWith('synonym')) { url += `synonym/${id}/` }
+    else if (id.startsWith('stopword')) { url += `stopword/${id}/` }
+    else { url += `entry/${id}/`; isEntry = true; }
```
nobeeakon, line 15: what about passing this logic to a hook or even just a function? That way you could call it where needed. In any case it may be better to at least pass this piece of logic to the `AccumulateAllComponents` as is not really used in here

taxonomy-editor-frontend/src/pages/editentry/index.jsx
aadarsh-ram: @nobeeakon That's just cool! I haven't heard of the `.?` operator. I'll add it and remove the variable.

taxonomy-editor-frontend/src/pages/editentry/index.jsx
aadarsh-ram: This logic is being used to find the url to which, the request has to be sent. My goal was to first retrieve the information in `editentry/index.jsx` and pass along the information to `AccumulateAllComponents.jsx`. I would be able to bring this logic to `AccumulateAllComponents` but I thought it would make the component even more cluttered and difficult to understand.

Regarding making it a function/hook, this logic will only be used here and will most probably not be reused anywhere else. @nobeeakon do you think there is any other way to structure it, without compromising the readability?

taxonomy-editor-frontend/src/pages/editentry/FetchRelations.jsx
```diff
@@ -0,0 +1,30 @@
+import useFetch from "../../components/useFetch";
+import { Typography } from "@mui/material";
+import { Link } from "react-router-dom";
+import OpenInNewIcon from '@mui/icons-material/OpenInNew';
+
+const FetchRelations = ({url, title}) => {
```
nobeeakon, line 6: this may be renamed, as is not only fetching but also displaying

taxonomy-editor-frontend/src/pages/editentry/index.jsx
nobeeakon: if not reused anywhere else think is ok to not doing a hook/function. I'd advice anyways to do it in the `AccumulateAllComponents` as that is the one using it, besides you need to pass the `url` to the `FetchRelations` so if you fetch the data at the `index` level you still need to have the `url` in there

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
```diff
@@ -1,19 +1,30 @@
 import { IconButton, TextField, Typography } from "@mui/material";
 import AddBoxIcon from '@mui/icons-material/AddBox';
-import { DataGrid } from '@mui/x-data-grid';
+import MaterialTable from '@material-table/core';
+import { useState } from "react";
 
 const ListAllProperties = ({ nodeObject, setNodeObject }) => {
     let toBeRendered = []
+
+    function guidGenerator() {
```
nobeeakon, line 9: don't understand what's the value of doing this?

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
```diff
@@ -38,23 +49,56 @@ const ListAllProperties = ({ nodeObject, setNodeObject }) => {
...
+                ]}
+                editable={{
+                onRowAdd: (newRow) => new Promise((resolve, reject) => {
+                    const updatedRows = [...data, { id: Math.floor(Math.random() * 100), ...newRow }]
+                    setTimeout(() => {
+                    setData(updatedRows)
+                    resolve()
+                    }, 2000)
+                }),
+                onRowDelete: selectedRow => new Promise((resolve, reject) => {
```
nobeeakon: mmm, from my understanding this may be better to keep it in a different PR as this one is already super big. Also, do you want to store those changes in the database?, if not I don't see a lot of value in having a delete update, etc

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
aadarsh-ram: It's to generate a unique key for each row of the properties table.

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
```diff
@@ -1,19 +1,30 @@
...
                 toBeRendered.push({
-                    'id': index+1,
+                    'id': guidGenerator(),
                     'property-name': property_name,
                     'property-value': nodeObject[key]
                 })
             }
     });
+    
+    const [data, setData] = useState(toBeRendered);
```
nobeeakon, line 28: better not call it this way, please prefer to have the hooks before anything else is being called, as they can't be after conditionals, etc. You could instead declare it in line `7` and set it later, possibly in a `useEffect`

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
```diff
@@ -38,23 +49,56 @@ const ListAllProperties = ({ nodeObject, setNodeObject }) => {
...
+                title={<Typography sx={{ml: 2, mt: 2, mb: 1}} variant='h5'>Properties</Typography>}
+                data={data}
+                columns={[
+                    { title: 'Name', field: 'property-name' },
+                    { title: 'Value', field: 'property-value' }
+                ]}
+                editable={{
+                onRowAdd: (newRow) => new Promise((resolve, reject) => {
+                    const updatedRows = [...data, { id: Math.floor(Math.random() * 100), ...newRow }]
+                    setTimeout(() => {
```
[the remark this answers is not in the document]
aadarsh-ram: Using a timeout for allowing it to update the state.

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
aadarsh-ram: @nobeeakon Yep, all the changes need to be tracked, so that I'll be able to update it in the database.

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
```diff
@@ -1,19 +1,30 @@
...
+    function guidGenerator() {
+        var S4 = function() {
+           return (((1+Math.random())*0x10000)|0).toString(16).substring(1);
+        };
+        return (S4()+S4()+"-"+S4()+"-"+S4()+"-"+S4()+"-"+S4()+S4()+S4());
+    }
+
     Object.keys(nodeObject).forEach((key, index) => {
         if (key.startsWith('prop')) {
                 let property_name = key.split('_').slice(1).join('_');
```
nobeeakon, line 19: please include in a comment how the key looks like. Otherwise it is difficult to follow why the split.slice.join

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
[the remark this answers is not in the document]
nobeeakon: I may be missing something, but as far as I can see, it is not needed

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
nobeeakon: still, please if possible, do not continue adding features in this PR

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
nobeeakon: thanks, I see, mmm,  if you need this, possibly the place to do it is when loading the data or inside a `useEffect` as proposed below, that way you know when it will run and why it will run. Also you would need to do it on every action (delete, add new item)...
(think [this](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) may help)

taxonomy-editor-frontend/src/pages/editentry/ListAllProperties.jsx
nobeeakon: amm, you may want to do the databse update and the client update in a single PR so you have the whole feature, as that may  affect the implementation

aadarsh-ram: Yes @nobeeakon, after successful implementation of CRUD operations in the table present in `ListAllProperties.jsx`, we can merge after your approval. You have also been invited to the taxonomy-editor team. Hence after accepting it, you would be able to merge PRs too.

taxonomy-editor-frontend/src/components/useFetch.jsx
alexgarel: I think this kind of suggestion is useful @nobeeakon 
introducing right libraries up front to avoid reinventing the wheel latter on ;-)

(but this is a difficult equilibrium to get unless you are experimented, so cool to see suggestions on that).

nobeeakon:
this PR was split in #42 to #47

PR #35, opened 2022-07-30
feat: Add GET paths for parents and children + fix: Update Neo4J query

### LINKED ISSUE #33: Add GET path for finding relationships of a node
There is no path for finding relationships (mainly `is_child_of` relationship) for any node. This must be added to compute the immediate parents and children of a particular node. (Will be helpful for the frontend)

### LINKED ISSUE #40: Update Neo4J query to disallow partial updates
Currently, the frontend tracks all changes (creation, addition, deletion) on keys and values of a node, and POSTs all information (such as translations, preceding_lines, properties etc) about a node. 

Since, the query only allows partial updates, deleted keys are not tracked in the database. Hence, the update db query has to be changed to allow full updates to a node.

aadarsh-ram:
### What
- Adds GET paths `/entry/{entry}/parents` and `/entry/{entry}/children` 
- Used for finding parents and children of a node
- Changed neo4j query to disallow partial updates

### Related issue(s)
- Fixes #33 
- Fixes #40

### DIFF
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -75,7 +75,7 @@ def update_nodes(label, entry, incomingData):
             raise ValueError("Invalid key: %s", key)
     
     # Build query
-    query = [f"""MATCH (n:{label}) WHERE n.id = $id"""]
+    query = [f"""MATCH (n:{label}) WHERE n.id = $id """, """SET n={} """, """SET n.id = $id"""]
 
     for key in incomingData.keys():
         query.append(f"""\nSET n.{key} = ${key}\n""")
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -3,7 +3,7 @@ Database helper functions for API
 """
 import re
 from neo4j import GraphDatabase         # Interface with Neo4J
-from . import settings                         # Neo4J settings
+from . import settings                  # Neo4J settings
 
 def initialize_db():
     """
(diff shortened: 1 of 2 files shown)


backend/editor/entries.py
```diff
@@ -53,7 +75,7 @@ def update_nodes(label, entry, incomingData):
             raise ValueError("Invalid key: %s", key)
     
     # Build query
-    query = [f"""MATCH (n:{label}) WHERE n.id = $id"""]
+    query = [f"""MATCH (n:{label}) WHERE n.id = $id """, """SET n={} """, """SET n.id = $id"""]
```
alexgarel, line 78: The idea here is that your reset the node before running updates ?

alexgarel (changes requested):
It's ok but I have one question:

> * Changed neo4j query to disallow partial updates
Why do we want this ? Shan't it be a parameter or a specific url ?

Also if we have to push every key / values each time, we should refactor the code, because it can be done in only one query instead of so many.

backend/editor/entries.py
aadarsh-ram: Yess. The React App sends all the current keys of a node and the backend updates only those keys. But, if a user deletes any key, the change doesn't reflect. Hence, I clear the node and start adding all the keys sent by the React App.

aadarsh-ram: If there's a better way to approach this, do let me know.

alexgarel: Hi @aadarsh-ram, the way to remove a property could be to just set it to `null`

So if I want to remove the "tags_fr" key, I send "tags_fr: null".

aadarsh-ram:
@alexgarel I have now figured out a new way to tackle this. I fetch the current node information and compare it with the node object received, to remove the deleted keys. Does this make sense?

backend/editor/entries.py
```diff
@@ -74,9 +74,20 @@ def update_nodes(label, entry, incomingData):
         if not re.match(r"^\w+$", key) or key == "id":
             raise ValueError("Invalid key: %s", key)
     
+    # Get current node information and deleted keys
+    curr_node = list(get_nodes(label, entry).data()[0]['n'].keys())
+    deleted_keys = list(set(curr_node) ^ set(incomingData))
```
alexgarel, line 79: Don't you want to just remove keys where incoming data is null ?

alexgarel (approved):
I made a comment, but this can be ok, also like this. So fix it if you want.

PR #36, opened 2022-08-02
feat: Create unparser.py

BryanH01:
### What
- Script to write a .txt file from the nodes in neo4j database

### DIFF
--- a/parser/openfoodfacts_taxonomy_parser/unparser.py
+++ b/parser/openfoodfacts_taxonomy_parser/unparser.py
@@ -1,0 +1,125 @@
+from neo4j import GraphDatabase
+
+
+class WriteFile:
+    
+    def __init__(self, uri="bolt://localhost:7687"):
+        self.driver = GraphDatabase.driver(uri)
+        self.session = (
+            self.driver.session()
+        )  # Doesn't create error even if there is no active database
+    
+    def normalized_filename(self, filename):
+        """add the .txt extension if it is missing in the filename"""
+        return filename + (
+            ".txt" if (len(filename) < 4 or filename[-4:] != ".txt") else ""
+        )
+
+    def get_all_nodes(self):
+        """query the database and yield each node with its parents,
+            this function use the relationships between nodes"""
+        query = """
+            MATCH path = ShortestPath( (h:TEXT{id:"__header__"})-[:is_before*]->(f:TEXT{id:"__footer__"}) )
+            UNWIND nodes(path) AS n
+            RETURN n , [(n)-[:is_child_of]->(m) | m ]
+        """
+        results=self.session.run(query)
+        for result in results:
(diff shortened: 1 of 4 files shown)


parser/openfoodfacts_taxonomy_parser/unparser.py
```diff
@@ -0,0 +1,124 @@
+from neo4j import GraphDatabase
+
+
+class WriteFile:
```
alexgarel, line 4: Not a very descriptive name !

parser/openfoodfacts_taxonomy_parser/unparser.py
```diff
@@ -0,0 +1,124 @@
...
+        return lc_list
+
+    def write_tags_line(self, node, lc):
+        """return a string that should look like the original line"""
+        line = (", ").join(node["tags_" + lc])
+        line = lc + ":" + line + "\n"
+        return line
+
+    def write_comment_lines(self, comments):
+        """return a string that should look like the original comment lines"""
```
alexgarel, line 54: We should remove the `write_` to all those functions as they are not writting (and that's a good idea to do it like this)

parser/openfoodfacts_taxonomy_parser/unparser.py
```diff
@@ -0,0 +1,124 @@
...
+        """return a ordered list of properties with their language code (lc)"""
+        prop_list = []
+        for property in node:
+            if property.startswith("prop_"):
+                prop_list.append(property[5:])
+
+        # there is no rule for the order of properties
+        # properties will be arranged in alphabetical order
+        prop_list.sort()
+        return prop_list
```
alexgarel, line 69: typically done with a list comprehension:

```suggestion
        # there is no rule for the order of properties
        # properties will be arranged in alphabetical order
        return sorted(
            property[5:] for property in node if property.startswith("prop_")
        )
```

parser/openfoodfacts_taxonomy_parser/unparser.py
```diff
@@ -0,0 +1,124 @@
...
+        line = property_name + ":" + lc + ":" + property_value + "\n"
+        return line
+
+    def rewrite_file(self, filename):
+        """Write a .txt file with the given name"""
+        filename = self.normalized_filename(filename)
+        nodes = self.get_all_nodes()
+        previous_block = ""
+        with open(filename, "w", encoding="utf8") as file:
+            for node, parents in nodes:
```
alexgarel, line 84: there is a lot of logic here so it betters goes in a function or an iterator to always isolate write operations (always isolate code with side effects)

alexgarel (changes requested):
It's mostly ok, but I will do a quick pass to enhance readability.

Also we should have a test for this !

alexgarel (approved):

PR #39, opened 2022-08-05
feat: add checks

BryanH01:
On taxonomy:
- number of root entries
- list of root entries
- list of the 50 entries with the greatest number of direct children and their number of children

On an entry:
- list of translations present on children but not on the entry
- shortest and longest path to a root : path en length
- number of descendants
- list of properties that are defined on a parent
- list of properties that are already defined on a parent and redefined on the entry
- list of children than redefine one of the property of this entry
- number of children that use a synonym instead of a main id to link to this parent

For a language,
create a “languages” node:
- % of translated entries with at least 3 languages
- list of entries that miss translation (<3)
storing results in sub-node for each language:
- % of translated entries

### DIFF
--- a/parser/openfoodfacts_taxonomy_parser/parser.py
+++ b/parser/openfoodfacts_taxonomy_parser/parser.py
@@ -340,6 +340,223 @@ class Parser:
         self.create_previous_link()
         # self.delete_used_properties()
 
+    def check(self):
+        #check taxonomy
+        self.check_roots()
+        self.check_big_parents()
+
+        #check entries
+        self.check_shortest_and_longest_path()
+        self.check_number_of_descendants()
+        self.check_parents_prop()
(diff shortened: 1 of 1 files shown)


alexgarel:
@BryanH01 there has been a misunderstanding here (sorry for that).

I didn't want you to add the checks at parsing time, nor statically.

What should be done is a checks.py file that has all those functions (and maybe one that run all checks and report).

The idea is that after modifications of the taxonomy, @aadarsh-ram code would call those checks to eventually report errors to the user: either on a "warning / error pages" or on single entries (depending on the check).

alexgarel:
@teolemon this PR is stalled and it does not do the checks the way we want them (we a live API), we keep it just for the neo4j request that can be re-used.

cc: @aadarsh-ram

PR #41, opened 2022-08-12
feat: Add CRUD features for entries and root nodes

### LINKED ISSUE #34: React frontend for updation and retrieval of entries

aadarsh-ram:
### What
Implemented CRUD operations for:
- Properties of an entry (Editable table)
- Translations of an entry
- Children of an entry
- Root nodes

### Related issue(s)
- Fixes #34 

### Screenshot
Properties table
Edit translations

(Other screenshots will be added)


taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -14,46 +15,104 @@ import ListAllEntryRelationships from "./ListAllEntryRelationships";
...
             }
+            {/* Button for submitting edits */}
+            <Button
+                variant="contained"
+                onClick={handleSubmit}
+                sx={{ml: 4, mt:2, width: 130}}>
+                    Submit
+            </Button>
+            {/* Dialog box for acknowledgment of update */}
+            <Dialog
```
nobeeakon, line 93: what about using a snackbar: https://mui.com/material-ui/react-snackbar/#customization instead of the Dialog, I feel is a simpler way of notifying the user

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -14,46 +15,104 @@ import ListAllEntryRelationships from "./ListAllEntryRelationships";
...
+                    headers: {"Content-Type" : "application/json"},
+                    body: JSON.stringify(updateChildren)
+                }).then(() => {
+                    setOpen(true);
+                }).catch((errorMessage) => {
+                    console.log(errorMessage);
+                })
+            }
+        }).catch((errorMessage) => {
+            console.log(errorMessage);
```
nobeeakon, line 70: no need to console log, think it is ok if leave it as `catch(()=> {})`

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -14,46 +15,104 @@ import ListAllEntryRelationships from "./ListAllEntryRelationships";
...
+    }
+    
+    // Helper functions for Dialog component 
+    const handleClose = () => {setOpen(false)};
+    const handleBack = () => {navigate('/entry')};
+
+    // Function handling updation of node
+    const handleSubmit = () => {
+        const {id, ...data} = updatedNodeObject // ID not allowed in POST
+        fetch(url, {
```
nobeeakon, line 53: would it make sense to use `Promise.all` that way you can handle a single catch and a single success (`setOpen(true)`)

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -14,46 +15,104 @@ import ListAllEntryRelationships from "./ListAllEntryRelationships";
...
+                <DialogTitle id="alert-dialog-title">
+                {"Your edits have been saved!"}
+                </DialogTitle>
+                <DialogContent>
+                <DialogContentText id="alert-dialog-description">
+                    The node {id} has been successfully updated.
+                </DialogContentText>
+                </DialogContent>
+                <DialogActions>
+                <Button sx={{fontFamily: 'Roboto, Helvetica, Arial, sans-serif'}} onClick={handleBack}>
```
nobeeakon, line 107: instead of navigate, better use a `Link`

nobeeakon: why these have a different font family?

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -14,46 +15,104 @@ import ListAllEntryRelationships from "./ListAllEntryRelationships";
...
+    const [nodeObject, setNodeObject] = useState(null); // Storing original node information
+    const [updatedNodeObject, setUpdatedNodeObject] = useState(null); // Storing updates to node
+    const [updateChildren, setUpdateChildren] = useState([]); // Storing updates of children in node
+    const [open, setOpen] = useState(false); // Used for Dialog component
+    const navigate = useNavigate(); // Navigation between pages
     const { data: node, isPending, isError, isSuccess, errorMessage } = useFetch(url);
 
     // Setting state of node after fetch
     useEffect(() => {
         setNodeObject(node?.[0]);
```
nobeeakon, line 32: does this needs to be in a state? what about using directly as `const originalNode = node?.[0]`

taxonomy-editor-frontend/src/pages/editentry/AccumulateEntryInfo.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Parent component used for rendering sub-components
+ * Sub-components are used to render info for an "entry"
+ */
+
+const AccumulateEntryInfo = ({ nodeObject, setNodeObject, originalNodeObject }) => {
+    
+    return (
+        <Box className="allEntryProperties">
```
nobeeakon, line 13: please remove the `className`

taxonomy-editor-frontend/src/pages/editentry/AccumulateEntryInfo.jsx
```diff
@@ -0,0 +1,22 @@
+import { Box, Typography } from "@mui/material";
+import ListAllEntryProperties from "./ListAllEntryProperties";
+import ListTranslations from "./ListTranslations";
+
+/**
+ * Parent component used for rendering sub-components
+ * Sub-components are used to render info for an "entry"
+ */
+
+const AccumulateEntryInfo = ({ nodeObject, setNodeObject, originalNodeObject }) => {
```
nobeeakon, line 10: maybe this component could be used directly, I mean, remove this component and pass the things to the parent

taxonomy-editor-frontend/src/pages/editentry/ListAllEntryProperties.jsx
```diff
@@ -1,68 +1,138 @@
...
+        setNodeObject(prevState => {
+            const toRemove = "prop_"+key;
+            const {[toRemove]: _, ...newNodeObject} = prevState;
+            return newNodeObject
+        })
     }
     
     return (
-        <Box>
+        <Box className="all-properties">
```
nobeeakon, line 58: please remove `className`

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
aadarsh-ram: Since this is a priority notification, I thought it should have a dialog of its own. Having a snackbar for acknowledgement wouldn't be ideal from a user's point of view.

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
aadarsh-ram: The button text wasn't clear. I'll add it to the ThemeProvider.

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
aadarsh-ram: Yep, it needs to be in a state. Since if assignment happens inside the useEffect, that variable's scope isn't global. If it was defined in a 'let', the contents will be lost after each render.

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -45,42 +45,36 @@ const AccumulateAllComponents = ({ id }) => {
...
+        }
+        Promise.all(allUrlsAndData.map(([url, data]) => {
+            return fetch(url, {
+                method : 'POST',
+                headers: {"Content-Type" : "application/json"},
+                body: JSON.stringify(data)
+            }).then(() => {
+                setOpen(true);
+            }).catch(() => {})
+        }))
```
nobeeakon, line 63: nit: 
```
Promise.all(allUrlsAndData.map(([url, data]) => {
    return fetch(url, {
        method : 'POST',
        headers: {"Content-Type" : "application/json"},
        body: JSON.stringify(data)
    })
})).then(() => {
    setOpen(true);
}).catch(() => {})
```

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
aadarsh-ram: Yep, changed it to a derived state.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,353 @@
...
 /**
  * Sub-component for rendering translation of an "entry"  
 */
-const ListTranslations = ({ nodeObject, setNodeObject }) => {
-    let renderedTranslations = {}
+const ListTranslations = ({ nodeObject, setNodeObject, originalNodeObject }) => {
 
-    Object.keys(nodeObject).forEach((key) => {
+    let [renderedTranslations, setRenderedTranslations] = useState([]) // Stores state of all tags
+    let [mainLang_renderedTranslations, setMainLang_renderedTranslations] = useState([]) // Stores state of main language's tags
```
nobeeakon, line 19: nit: better to have these as `const`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,353 @@
...
+                    <AddBoxIcon />
+                </IconButton>
+            </Stack>
+            
             {/* Render main language tags */}
             <Typography variant='h6'> 
-                {
-                    nodeObject["tags_"+nodeObject['main_language']].map((tag, index) => {
+                { nodeObject && 
+                    mainLang_renderedTranslations.map((tagObj) => {
```
nobeeakon, line 247: nit: `mainLang_renderedTranslations.map((tagObj) => {` could be `mainLang_renderedTranslations.map(({index, tag}) => {`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,353 @@
...
+                                    <TextField 
+                                        size="small" 
+                                        sx={{mt: 1}} 
+                                        onChange = {event => {
+                                            changeData(nodeObject['main_language'], index, event.target.value)
+                                        }}
+                                        value={tag}
+                                        variant="outlined" />  
+                                </Paper>
+                                    <IconButton sx={{ml: 1, mt: 1, color: "#808080"}} onClick={(e) => handleDelete(nodeObject.main_language, index, e)}>
```
nobeeakon, line 262: nit: no need to pass the `e` in the `onClick`, can be `onClick={() => handle...}`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+        Object.keys(originalNodeObject).forEach((key) => {
+        
+            // Get all tags and its corresponding language code
+            // Tagids need to be recomputed, so shouldn't be rendered
+            // Eg: tags_fr
+        
+            if (key.startsWith('tags') && !key.includes('ids')) {
+
+                // If tags are not for the main language, add them to otherLangTags
+                if (!key.endsWith(originalNodeObject.main_language)) {
```
nobeeakon, line 76: nit: `if(key.includes(originalNodeObject.main_language)) {...// main language} else { //not main language}`  it is usually easier to follow the `if (contidion === true) {}` than `if (negate condition) {} else { // which would be the condition === true anyways}`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+        setOpen(false);
+    }
+
+    // Changes the translations to be rendered
+    // Dependent on changes occuring in "originalNodeObject"
+    useEffect(() => {
+
+        // Main langauge tags are considered separately, since they have to be rendered first
+        let mainLangTags = []
+        let otherLangTags = []
```
nobeeakon, line 65: nit: no need to be `let`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                // If tags are not for the main language, add them to otherLangTags
+                if (!key.endsWith(originalNodeObject.main_language)) {
 
-                // Slice the language code
-                let languageCode = key.slice(-2);
-                renderedTranslations[languageCode] = nodeObject[key]
+                    // Slice the language code
+                    let languageCode = key.slice(-2);
+                    let tobeInsertedObj = {'languageCode' : languageCode, 'tags' : []} // General format for storing tags for different lc's
+                    originalNodeObject[key].map((tag) => (
```
nobeeakon, line 81: nit: better to not mutate inside a `.map`, this could be `.forEach`. But better do no mutation at all, it can lead to bugs or unexpected behaviors. 
```
const newNode = {...originalNodeObject}
const tags = newNode[key].map(tag => { ... })
```
But... I don't see where the originalNodeObject is being set again ... is that part missing and we are rellying in the mutation?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                        tobeInsertedObj['tags'].push({
+                            'index' : uuid.v4(), // Give a unique identifier for each tag
```
nobeeakon, line 83: why would we want to create new uuids? even for tags that already exist in the `originalNodeObject`?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                    ))
+                }
             }
-    })
-    
+        })
+
+        // Set states
+        setMainLang_renderedTranslations(mainLangTags);
+        setRenderedTranslations(otherLangTags);
```
nobeeakon, line 103: here we are setting these two states, but both are directly related to the `originalNodeObject` could we remove these two and directly use the `originalNodeObject` and take these values as a derived state instead?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                            'tag' : tag
+                        })
+                    ))
+                    otherLangTags.push(tobeInsertedObj);
+                }
+                // If tags are for main language, add them to mainLangTags
+                else {
+                    originalNodeObject["tags_"+originalNodeObject['main_language']].map((tag) => (
+                        mainLangTags.push({ 
+                            'index' : uuid.v4(),
```
nobeeakon, line 93: same comment on reseting the uuids

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
 /**
  * Sub-component for rendering translation of an "entry"  
 */
-const ListTranslations = ({ nodeObject, setNodeObject }) => {
-    let renderedTranslations = {}
+const ListTranslations = ({ nodeObject, setNodeObject, originalNodeObject }) => {
 
-    Object.keys(nodeObject).forEach((key) => {
+    const [renderedTranslations, setRenderedTranslations] = useState([]) // Stores state of all tags
+    const [mainLang_renderedTranslations, setMainLang_renderedTranslations] = useState([]) // Stores state of main language's tags
```
nobeeakon, line 19: nit: let's keep the naming consistent, think in general is using camelCase

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+    const [open, setOpen] = useState(false); // Used for Dialog component
+    const [newLanguageCode, setNewLanguageCode] = useState(''); // Used for storing new LC from Dialog
+    const [isValidLanguageCode, setisValidLanguageCode] = useState(false); // Used for validating a new LC
```
nobeeakon, line 22: nit: let's keep the naming consistent, think in general is using camelCase

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                    autoFocus
+                    margin="dense"
+                    onKeyPress={(e) => { (e.key === 'Enter') && isValidLanguageCode && handleAddTranslation(newLanguageCode, e) }} 
+                    onChange={(e) => { 
+                        setNewLanguageCode(e.target.value);
+                        const validateBool = ISO6391.validate(e.target.value);
+                        const ifDuplicateBool = renderedTranslations.some(el => (el.languageCode === e.target.value)) || 
+                                                nodeObject.main_language === e.target.value
+                        validateBool && !ifDuplicateBool ? setisValidLanguageCode(true) : setisValidLanguageCode(false)
+                        validateBool && !ifDuplicateBool ? setBtnDisabled(false) : setBtnDisabled(true)
```
nobeeakon, line 332: this could be a function?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+        if (key === nodeObject.main_language) {
+            const duplicateMainLang_renderedTranslations = mainLang_renderedTranslations.filter(obj => !(index === obj.index))
+            setMainLang_renderedTranslations(duplicateMainLang_renderedTranslations); // Set state
+
+            // Updated tags assigned for later use
+            tagsToBeInserted = duplicateMainLang_renderedTranslations.map(el => (el.tag))
+        }
+        // State of "renderedTranslations" is updated according to format used
+        else {
+            let newRenderedTranslations = []
```
nobeeakon, line 199: no need to be `let`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                    duplicateOtherTags.push(allTagsObj)
+                }
+            })
+            // Set state
+            setRenderedTranslations(duplicateOtherTags)
+        }
+
+        let tagsToBeInserted = updatedTags.map(el => (el.tag)) // Removes unique idenitifer from each tag
+
+        key = 'tags_' + key; // LC must have a prefix "tags_"
```
nobeeakon, line 145: nit: carefull here, you are reseting something that comes from the outside

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
  * Sub-component for rendering translation of an "entry"  
 */
-const ListTranslations = ({ nodeObject, setNodeObject }) => {
-    let renderedTranslations = {}
+const ListTranslations = ({ nodeObject, setNodeObject, originalNodeObject }) => {
 
-    Object.keys(nodeObject).forEach((key) => {
+    const [renderedTranslations, setRenderedTranslations] = useState([]) // Stores state of all tags
+    const [mainLang_renderedTranslations, setMainLang_renderedTranslations] = useState([]) // Stores state of main language's tags
+    const [open, setOpen] = useState(false); // Used for Dialog component
```
nobeeakon, line 20: better naming for this?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
nobeeakon: also you are using ternaries in a non standar  way, those are mainly for returning something.
 `if (validateBool && !ifDuplicateBool ) {setisValidLanguageCode(true);setBtnDisabled(false)} else {...}`

nobeeakon: could the `btnDisabled` be `!isValidLanguageCode` ?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
     }
 
     return ( 
-        <Box sx={{ml: 4}} className="translations">
+        <Box sx={{ml: 4}}>
             {/* Title */}
-            <Typography sx={{mt: 4, mb: 1}} variant='h5'>Translations</Typography>
+            <Stack direction="row" alignItems="center">
+                <Typography sx={{mt: 4, mb: 1}} variant='h5'>Translations</Typography>
+                <IconButton sx={{mt: 3.5, ml: 1, color: "#808080"}} onClick={handleOpen}>
```
nobeeakon, line 229: same comment with using custom colors. Please avoid that, if not possible to set it in the ThemeProvider, then at leas have those in a `const customColors = { }` somewhere so can be reused or easily tracked

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+
             {/* Main Language */}
-            <Typography variant='h6'>
-                { ISO6391.getName(nodeObject.main_language) }
-            </Typography>
+            <Stack direction="row" alignItems="center">
+                <Typography variant='h6'>
+                    { nodeObject && ISO6391.getName(nodeObject.main_language) }
+                </Typography>
+                <IconButton sx={{ml: 1, color: "#808080"}} onClick={() => handleAdd(nodeObject.main_language)}>
```
nobeeakon, line 239: please avoid custom colors, see comment above

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
                     })
                 }
             </Typography>
 
             {/* All other languages */}
             {
-                Object.entries(renderedTranslations).map( ([lang, value]) => {
+                renderedTranslations.map( (allTagsObj) => {
+                    const lang = allTagsObj['languageCode']
+                    const value = allTagsObj['tags']
```
nobeeakon, line 273: better naming for `value`, that's super generic

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                                                <TextField 
+                                                    size="small" 
+                                                    sx={{mt: 1}} 
+                                                    onChange = {event => {
+                                                        changeData(lang, index, event.target.value)
+                                                    }}
+                                                    value={tag} 
+                                                    variant="outlined" />
+                                            </Paper>
+                                            <IconButton sx={{ml: 1, mt: 1, color: "#808080"}} onClick={() => handleDelete(lang, index)}>
```
nobeeakon, line 304: same comment on custom colors

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
-                        <Box key={lang} className="translation-component">
-                            <Typography sx={{mt:1}} variant="h6">
-                                {ISO6391.getName(lang)}
-                            </Typography>
+                        <Box key={lang}>
+                            <Stack sx={{mt: 2}} direction="row" alignItems="center">
+                                <Typography variant="h6">
+                                    {ISO6391.getName(lang)}
+                                </Typography>
+                                <IconButton sx={{ml: 1, color: "#808080"}} onClick={() => handleAdd(lang)}>
```
nobeeakon, line 280: same comment on custom colors

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                                    <AddBoxIcon />
+                                </IconButton>
+                                <IconButton sx={{ml: -1, color: "#808080"}} onClick={() => handleDeleteTranslation(lang)}>
```
nobeeakon, line 283: same comment on custom colors

nobeeakon: is it really necessary to have `ml: -1`?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+            <Dialog open={open} onClose={handleClose}>
+                <DialogTitle>Add a language</DialogTitle>
+                <DialogContent>
+                <DialogContentText>
+                    Enter the two letter language code for the language to be added.
+                </DialogContentText>
+                <TextField
+                    autoFocus
+                    margin="dense"
+                    onKeyPress={(e) => { (e.key === 'Enter') && isValidLanguageCode && handleAddTranslation(newLanguageCode, e) }}
```
nobeeakon, line 325: maybe better to use `charCode`, please check https://caniuse.com/?search=event.code

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                                    <TextField 
+                                        size="small" 
+                                        sx={{mt: 1}} 
+                                        onChange = {event => {
+                                            changeData(nodeObject['main_language'], index, event.target.value)
+                                        }}
+                                        value={tag}
+                                        variant="outlined" />  
+                                </Paper>
+                                    <IconButton sx={{ml: 1, mt: 1, color: "#808080"}} onClick={() => handleDelete(nodeObject.main_language, index)}>
```
nobeeakon, line 260: same comment on color

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
-                                    size="small" 
-                                    sx={{mt: 1}} 
-                                    onChange = {event => {
-                                        changeData(nodeObject['main_language'], index, event.target.value)
-                                    }}
-                                    value={tag} 
-                                    variant="outlined" />
-                            </Paper>
+                            <Stack key={index} direction="row" alignItems="center">
+                                <Paper component={Stack} direction="column" sx={{ml: 4, width: 200}}>
```
nobeeakon, line 250: here you have a Stack inside another Stack. Are those needed? I wuold expect that TextField can be side by side with a button. Also is there any reason for that measure in the `width`?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                <Typography variant='h6'>
+                    { nodeObject && ISO6391.getName(nodeObject.main_language) }
+                </Typography>
+                <IconButton sx={{ml: 1, color: "#808080"}} onClick={() => handleAdd(nodeObject.main_language)}>
+                    <AddBoxIcon />
+                </IconButton>
+            </Stack>
+            
             {/* Render main language tags */}
             <Typography variant='h6'>
```
nobeeakon, line 245: do we really want h6 typography into this mapped elements? why not the usual default body typography?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
aadarsh-ram: Since I'm pushing values to it, it has been set as a variable.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
aadarsh-ram: `originalNodeObject` is the result of the node fetched. It changes only when a new node is fetched from the backend API. 

The tags related to main_language and other languages are stored in two separate arrays `mainLangTags` and `otherLangTags`.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
aadarsh-ram: We create new uuids only when a new node is being fetched. As `originalNodeObject` will never be altered by "non-fetch" code, it has wrapped inside a `useEffect`.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
nobeeakon: yep, no need to be `let`. In Js, `let ` is used for reseting the variable. Eg: 
```
let myVar = [1,2]
myVar = [2,3]
```
In the case of `const`, is basically to define a reference to a value and not reset that reference
```
const myConst = []
myConst = [] // this will throw an error
```
but... ` myConst.push[1]` won't throw an error, as you are not reseting to another value but modifying the array itself. This only applies for `[]` or `{}`, not to numbers, booleans, etc.
 
Although you can use them interchangably...

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
nobeeakon: anyways, if using `.map()` better to have the usual syntax `const something = arr.map( ...)`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
nobeeakon: well yea, I mean, I was not saying that the `originalNodeObject` would be updated, but that you are mapping (`originalNodeObject[key].map((tag) => {})`) values from the `originalNodeObject` and assigning them a new `index`. I'd assume that those  already have an `index` so is it possible to use the one they already have? On the other hand I'd expect those `index` to come (and be used) from the database, if that's not the case why even bother on defining those in the client, but if that's actually the case why defining them in the client and modifying the one they should already have?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+                }
             }
-    })
-    
+        })
+
+        // Set states
+        setMainLang_renderedTranslations(mainLangTags);
+        setRenderedTranslations(otherLangTags);
+    }, [originalNodeObject]);
```
nobeeakon, line 104: Please check the comment above on this but, I was wondering what's the need for multiple states derived one from the other? and if it may be better to only use the `nodeObject` as it is a state and not a derived value?
This comment applies in the other `useEffect`s that have `originalNodeObject` as a dependency (in this file and in `ListAllEntryProperties`).

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -1,94 +1,351 @@
...
+    }
+
+    // Used for deleting a translation language
+    function handleDeleteTranslation(key) {
+        const newRenderedTranslations = renderedTranslations.filter(obj => !(key === obj.languageCode))
+        setRenderedTranslations(newRenderedTranslations);
+        key = 'tags_' + key; // LC must have a prefix "tags_"
+        
+        // Make changes to the parent NodeObject
+        setNodeObject(prevState => {
```
nobeeakon, line 51: related with other comments, but here you are defining two related values: `setRenderedTranslations` live in this component, `setNodeObject` lives in the parent. Both seem to hold the same information, why not using the one in the parent?  I'm wondering if this here is actually duplicating the info

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -14,46 +15,78 @@ import ListAllEntryRelationships from "./ListAllEntryRelationships";
...
-            </Box>
-        )
+        return (<Typography sx={{ml: 4}} variant='h5'>Loading..</Typography>)
+    }
+    
+    // Helper functions for Dialog component 
+    const handleClose = () => {setOpen(false)};
+
+    // Function handling updation of node
+    const handleSubmit = () => {
```
nobeeakon, line 47: please add a check `if (!nodeObject) return`

nobeeakon (approved):

PR #42, opened 2022-08-14
feat: Add home screen

aadarsh-ram:
### What
- Initialize frontend using Create-React-App
- Add home screen component

### Screenshot
home-screen


taxonomy-editor-frontend/src/App.js
```diff
@@ -0,0 +1,26 @@
...
+  typography: {
+      fontFamily : 'Plus Jakarta Sans',
+  },
+});
+
+function App() {
+  return (
+    <Router>
+      <ThemeProvider theme={theme}>
+        <div className="App">
```
nobeeakon, line 15: is this needed? Also for the order, think that the ThemeProvider and CssBaseline could go outside of the Router

taxonomy-editor-frontend/src/components/useFetch.jsx
```diff
@@ -0,0 +1,36 @@
+import { useState, useEffect } from "react";
+
+const useFetch = (url) => {    
+    const [data, setData] = useState(null)
+    const [isPending, setIsPending] = useState(true)
+    const [error, setError] = useState(null);
```
nobeeakon, line 6: these two (isPending, error) could be a variable called status, so the 'status' can be a string: 'error', 'pending', etc. Even better to have a single state that contains both the data and the status

taxonomy-editor-frontend/src/components/useFetch.jsx
```diff
@@ -0,0 +1,36 @@
...
+                setData(data);
+                setIsPending(false);
+                setError(null);
+            })
+            .catch(err => {
+                if (err.name === 'AbortError') {
+                    console.log("Fetch aborted!");
+                } else {
+                    setIsPending(false);
+                    console.log(err);
```
nobeeakon, line 27: logs are good during development, but in general not really useful to deliver them to production

taxonomy-editor-frontend/src/pages/home/index.jsx
```diff
@@ -0,0 +1,63 @@
+import Button from '@mui/material/Button';
+import {Link as MuiLink} from '@mui/material';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import Container from '@mui/material/Container';
+import { Link } from "react-router-dom";
+
+function Copyright() {
```
nobeeakon, line 8: If not resused, this could be part of the code, no need to have an extra component

taxonomy-editor-frontend/src/pages/home/index.jsx
```diff
@@ -0,0 +1,63 @@
...
+import {Link as MuiLink} from '@mui/material';
+import Box from '@mui/material/Box';
+import Typography from '@mui/material/Typography';
+import Container from '@mui/material/Container';
+import { Link } from "react-router-dom";
+
+function Copyright() {
+  return (
+    <Typography variant="body2" color="text.secondary" align="center" sx={{ mt: 8, mb: 4 }}>
+      {'Copyright © '}
```
nobeeakon, line 11: nit: this text doesn't needs to be inside {''}

taxonomy-editor-frontend/src/components/useFetch.jsx
aadarsh-ram: @nobeeakon I did most of my development using these variables separately. If I change it here, I would need refactor in a lot of places. I think it might be better to keep them separately instead, cause I would anyways make it into separate variables wherever I use them. Do you agree?

nobeeakon: mmm well, yeah I see the point, I think that changes are somehow inevitable when reviewing. Although I think it may work this way, and not modifying it would make us move faster (less things to revisit in the next commits),  think that having everything in a single state and having a `status` instead of `booleans` makes it easier to understand the logic, for example, in this case, you may need to always set three independent states. Lets say you first have an error and then the url changes and you have a success, currently you would have an error message from a previous request and a piece of data. So, if, for example in the component you first check if there's an error and display it, you could end up always showing it. As said, I think if we keep it like this it could work, but think it is good to keep things as clean/clear as we can. Also, I guess `error` could be renamed to `errorMessage`

aadarsh-ram: Yeah, I can rename that. Also, I'm not sure that would happen since the state variables displayed are totally dependent on the hook and if the hook runs, the states get reset. Is my intuition right?

taxonomy-editor-frontend/src/pages/home/index.jsx
aadarsh-ram: I'm using that convention to add a space after the copyright logo

taxonomy-editor-frontend/src/components/useFetch.jsx
nobeeakon: sorry @aadarsh-ram , not sure I understood :s

aadarsh-ram: > Lets say you first have an error and then the url changes and you have a success, currently you would have an error message from a previous request and a piece of data. So, if, for example in the component you first check if there's an error and display it, you could end up always showing it.

@nobeeakon In this case, all the states that are stored in the useFetch hook, are reset each time whenever the hook is called. So, if a URL changes, the useFetch hook will be called and previous states will be reset. So, I'll not be showing any previous errors if any. So, having separate variables won't result in any disadvantage I think. Does this make sense?

teolemon:
@aadarsh-ram can you add a few screenshots in the PR description ?

PR #43, opened 2022-08-14
feat: All entries page + Navbar

aadarsh-ram:
### What
- Completed an "All entries" page
- Added a Navbar component

### Screenshot
allentries-nav


taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
```diff
@@ -0,0 +1,72 @@
...
+import Paper from '@mui/material/Paper';
+import EditIcon from '@mui/icons-material/Edit';
+import IconButton from '@mui/material/IconButton';
+import { useNavigate } from "react-router-dom";
+
+const EntriesList = ({ nodes, title }) => {
+    const navigate = useNavigate();
+
+    // Handler function for button clicks
+    const handleClick = (event, id) => {
```
nobeeakon, line 17: what about using directly a `Link`?

taxonomy-editor-frontend/src/pages/allentries/EntriesList.jsx
```diff
@@ -0,0 +1,72 @@
...
+        event.preventDefault();
+        navigate('/entry/'+id);
+    }
+    return ( 
+        <div className="entry">
```
nobeeakon, line 22: please avoid using className, you could try `<Box sx={{}}>`. using hardcoded direct css classes tend to be difficult to manage  at some point :(

taxonomy-editor-frontend/src/pages/allentries/index.jsx
```diff
@@ -0,0 +1,22 @@
...
+        return (
+            <div className="all-entries">
+                {isError && <div>{errorMessage}</div>}
+            </div>
+        )
+    }
+    return (
+        <div className="all-entries">
+            {isPending && <div>Loading...</div>}
+            {isSuccess && <EntriesList nodes={nodes} title={"Test"}/>}
```
nobeeakon, line 17: nit: if there's no plan to include more things in this component, I think the content of `EntriesList` could live here (as is `EntriesList` is mainly `jsx` it should't be much of a problem)

taxonomy-editor-frontend/src/components/ResponsiveAppBar.jsx
```diff
@@ -0,0 +1,177 @@
...
+                width="50px"
+                height="50px"
+                alt="OpenFoodFacts logo"
+              />
+            </MuiLink>
+            <Typography
+              variant="h6"
+              noWrap
+              component="a"
+              href="/"
```
nobeeakon, line 142: maybe use `Link` instead?

taxonomy-editor-frontend/src/pages/allentries/index.jsx
```diff
@@ -1,20 +1,83 @@
...
+            <div className="all-entries">
+                {isPending && <div>Loading...</div>}
+            </div>
+        )
+    }
     return (
         <div className="all-entries">
-            {isPending && <div>Loading...</div>}
-            {isSuccess && <EntriesList nodes={nodes} title={"Test"}/>}
+            { isSuccess &&
```
nobeeakon, line 33: nit: one of the benefits of having an status vs different states in the fetch is that it is unsured that these different statuses (and its derived boolean  counterparts `isSuccess`, `isPending`, ...) are mutually exclusive, in this case as you have already checked that it is not loading or is not error then the only remaining option is `isSuccess`, so in principle there shouldn't be a need to check in here.

aadarsh-ram: Oh yeah, that makes sense. I'll remove that check.

nobeeakon (approved):
:+1:

PR #46, opened 2022-08-14
feat: Edit synonyms, stopwords page

aadarsh-ram:
### What
- Completed ListAllOtherProperties.jsx


taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -1,22 +1,97 @@
-import { Box, Typography } from "@mui/material";
+import { Box, Paper, Stack, TextField, Typography } from "@mui/material";
+import ISO6391 from 'iso-639-1';
 
 // Parent component used for rendering info
 // on a stopword, synonym, header or footer
```
nobeeakon, line 5: nit: if you use the other type of comment: `/** ...your comment here */` it is recognized as a doc string, please replace it

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -1,22 +1,97 @@
...
+
+    if (nodeObject) {
+        Object.keys(nodeObject).forEach((key) => {
+
+            // Get all tags and its corresponding language code
+            // Tagids need to be recomputed, so shouldn't be rendered
+
+            if (key.startsWith('tags') && 
+                !key.includes('ids')) {
+                    languageCode = key.slice(-2);
```
nobeeakon, line 22: mmm, maybe I'm missing something, but, it is redifining the `let languageCode` on every `key` of the `nodeObject`... it seems a bit unpredictable what would be the final value, makes me wonder if all have the same value and 1) what happens if so and 2) what happens if that's not the case ...

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
nobeeakon: what would be shown for header and footer? don't see anything related to those in the code... if not yet done please add a TODO, or remove it from the description

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
```diff
@@ -1,22 +1,97 @@
...
+                    <Typography sx={{ml: 8, mt: 1.5}} variant='h6'>
+                        {ISO6391.getName(languageCode)}
+                    </Typography>
+                </Box>
+                }
+            
+            {/* Stopwords or Synonyms */}
+            { Object.keys(renderedOtherProperties).length > 0 && 
+                <Box>
+                    {id.startsWith('stopword') ?
```
nobeeakon, line 71: .... made me think it it would be useful to have a function that returns what type of thing the `id` is, `getIdType(id)` and it would return `stopword` or `header` `footer`, etc. Could live in `createURL.js`. Also could be useful to conditionally render this component in `AccumulateAllComponents.jsx`. ... stretching it even more, for example the `isEntry` of `createUrl` function could be replaced to return the `type` obtained by that function and that way eventually used in other situations besides the check for if entry or not....

nobeeakon (changes requested):
nice work   🦅

taxonomy-editor-frontend/src/pages/editentry/ListAllOtherProperties.jsx
aadarsh-ram: @nobeeakon Yeah, you're right. I had actually thought about this, and wrapped the `Object.keys()` function inside a `useEffect` combined with state variables for `languageCode`. This will come in an another PR.

aadarsh-ram: I'll add a TODO to this to remind us that this will be changed in another PR.

taxonomy-editor-frontend/src/pages/editentry/createURL.jsx
```diff
@@ -4,15 +4,29 @@ export function createURL(id) {
...
+    else { url += `entry/${id}/` }
 
-    return {url, isEntry}
+    return url
+}
+
+export function getIdType(id) {
+
+    // Stores type of ID
+    // Eg: stopword, synonym, entry, header or footer
```
nobeeakon, line 22: Nice, I like the function 
nit: please turn this into a doc string for the function

taxonomy-editor-frontend/src/pages/editentry/ListAllNonEntryInfo.jsx
```diff
@@ -0,0 +1,103 @@
...
+    // Storing keys and values that needs to be rendered for editing
+    let renderedNonEntryInfo = {}
+
+    if (nodeObject) {
+        Object.keys(nodeObject).forEach((key) => {
+            /** 
+             * Get all tags and its corresponding language code
+             * Tagids need to be recomputed, so shouldn't be rendered
+             * Eg: tags_fr
+            */
```
nobeeakon, line 26: nit: this comment can be normal comment `//`, the benefit of having the doc string is for functions (components, etc) is that some IDEs show you the doc when using it

taxonomy-editor-frontend/src/pages/editentry/createURL.jsx
nobeeakon: now you could use it in the `createURL`

nobeeakon (approved):
nice work @aadarsh-ram

PR #59, opened 2022-08-19
ci: quality for parser

alexgarel:
### What
adding black, isort, flake8 and CI for parser


parser/tests/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+    session = driver.session()
+    connected = False
+    while not connected:
+        try:
+            session.run("MATCH () return 1 limit 1")
+        except ServiceUnavailable:
+            time.sleep(1)
+        else:
+            connected = True
+    return driver
```
alexgarel, line 22: @aadarsh-ram and @BryanH01 you might be interested in this pattern.
When making things run automatically it's one of the tricky part: waiting for third party service to be ready before running.

Here I made it a global fixture because that's a nice way to have it in the integrations tests and use it where needed.

teolemon (approved):

PR #74, opened 2022-09-07
fix: Error while updating a node (backend)

aadarsh-ram:
### What
- Accessing key-value pairs was implemented wrongly in the update function
- Two separate variables, one for keys and a dictionary for all data of the node has been created


alexgarel (approved):
LGTM

PR #76, opened 2022-09-11
feat: New paths for API

### LINKED ISSUE #38: Paths for updating relationships of a node
Paths must be created for updating parents and children of entries in a taxonomy. Creating these paths will allow the frontend send requests for updation.

### LINKED ISSUE #75: Paths for adding/delete nodes
New paths must be created for addition of root nodes and deletion of nodes in the backend API.

aadarsh-ram:
### What
- Adds GET paths for finding entry parents and children
- Adds POST path for updating entry children
- Adds POST paths for creation of new root node
- Adds DELETE path for deleting a node
- Some helper functions have also been created

### Related issue(s)
- Fixes #38 
- Fixes #75


backend/editor/entries.py
```diff
@@ -21,6 +21,91 @@ def shutdown_db():
...
+def create_node(label, entry, mainLanguageCode):
+    """
+    Helper function used for creating a node with given id and label
+    """
+    query = [f"""CREATE (n:{label})\n"""]
+
+    # Build all basic keys of a node
+    if (label == "ENTRY"):
+        canonicalTag = entry[3:]
+        query.append(f""" SET n.main_language = "{mainLanguageCode}" """) # Required for only an entry
```
alexgarel, line 42: We should use parameters as much as possible.
So here you should build a params dict and put mainLanguageCode inside. 
```python
query.append(f""" SET n.main_language = $main_language_code """)
params["main_language_code"] = main_language_code
```

alexgarel: same for canonical tag below.

Also we can't avoid the mainLanguageCode insertion in "n.tags_{mainLanguageCode}"

backend/editor/entries.py
```diff
@@ -21,6 +21,91 @@ def shutdown_db():
...
+def add_node_to_beginning(label, entry):
+    """
+    Helper function which adds an existing node to beginning of taxonomy
+    """
+    # Delete relationship between current first node and __header__
+    query = [f"""MATCH (b:TEXT)-[r:is_before]->(a) WHERE b.id = "__header__" DELETE r\n"""]
+    query.append(f"""RETURN a""")
+    result = session.run(" ".join(query))
+    end_node = result.data()[0]['a']
+    end_node_label = get_label(end_node['id']) # Get current first node ID
```
alexgarel, line 82: maybe we should rename start_node and start_node_label to be consistent ?

backend/editor/entries.py
```diff
@@ -21,6 +21,91 @@ def shutdown_db():
...
+
+    result = session.run(" ".join(query), {"id": entry})
+    return result
+
+def add_node_to_end(label, entry):
+    """
+    Helper function which adds an existing node to end of taxonomy
+    """
+    # Delete relationship between current last node and __footer__
+    query = [f"""MATCH (a)-[r:is_before]->(b:TEXT) WHERE b.id = "__footer__" DELETE r\n"""]
```
alexgarel, line 58: Here and below, I think we would gain in clarity by avoiding `a`,  `b` and `n`. (because it's a bit hard to read otherwise)

We could instead use `last_node`, `footer` and `new_node`.

backend/editor/entries.py
```diff
@@ -21,6 +21,91 @@ def shutdown_db():
...
+    query.append(f"""CREATE (a)-[:is_before]->(n)""")
+    result = session.run(" ".join(query), {"id": entry, "startnodeid": end_node['id']})
+
+def delete_node(label, entry):
+    """
+    Helper function used for deleting a node with given id and label
+    """
+    # Finding node to be deleted using node ID
+    query = []
+    query.append(f"""MATCH (n:{label})-[:is_before]->(a) WHERE n.id = $id\n""")
```
alexgarel, line 99: Here also renaming `a` and `b` to `next_node` and `previous_node` would help :-)

backend/editor/entries.py
```diff
@@ -101,4 +186,40 @@ def update_nodes(label, entry, incomingData):
...
-    return result
\ No newline at end of file
+    return result
+
+def update_node_children(entry, incomingData):
+    """
+    Helper function used for updation of node children with given id 
+    """
+    # Parse node ids from Neo4j Record object
+    current_children = [record["b.id"] for record in list(get_children(entry))]
```
alexgarel, line 196: What is this `b.id` parameter ? We should rename it `child.id`.

It's a change required in `get_children`.

It does not seems a big deal, but really naming variables and parameters rightly is very important in the long run.

backend/editor/entries.py
```diff
@@ -101,4 +186,40 @@ def update_nodes(label, entry, incomingData):
 
     params = dict(incomingData, id=entry)
     result = session.run(" ".join(query), params)
-    return result
\ No newline at end of file
+    return result
+
+def update_node_children(entry, incomingData):
```
alexgarel, line 191: `incomingData` is not descriptive enough here. If I read correcly this is the list of children id, so you should rename it `children_ids` or `new_ids`.

backend/editor/entries.py
```diff
@@ -101,4 +186,40 @@ def update_nodes(label, entry, incomingData):
...
+    Helper function used for updation of node children with given id 
+    """
+    # Parse node ids from Neo4j Record object
+    current_children = [record["b.id"] for record in list(get_children(entry))]
+    deleted_children = list(set(current_children) ^ set(incomingData))
+
+    # Delete relationships
+    for child in deleted_children:
+        query = []
+        main_language_code = child[:2]
```
alexgarel, line 202: Here again, use split instead of lenght of language code to stick to the spec.
```suggestion
        main_language_code = child.split(",", 1)[0]
```

backend/editor/entries.py
```diff
@@ -101,4 +186,40 @@ def update_nodes(label, entry, incomingData):
...
\ No newline at end of file
+    return result
+
+def update_node_children(entry, incomingData):
+    """
+    Helper function used for updation of node children with given id 
+    """
+    # Parse node ids from Neo4j Record object
+    current_children = [record["b.id"] for record in list(get_children(entry))]
+    deleted_children = list(set(current_children) ^ set(incomingData))
```
alexgarel, line 197: This is not clear, you are mixing deleted_children and added_children.

So really you should have this instead:
```suggestion
    deleted_children = set(current_children) - set(incomingData)
    added_children = set(incomingData) - set(current_children)
```
Then your code can treat them separately.

backend/editor/entries.py
```diff
@@ -101,4 +186,40 @@ def update_nodes(label, entry, incomingData):
...
+
+    # Delete relationships
+    for child in deleted_children:
+        query = []
+        main_language_code = child[:2]
+        is_old_node = list(get_nodes("ENTRY", child))
```
alexgarel, line 203: It would be better to check existence of all nodes in only one request  and then again to get non_existing nodes by comparing new ids and ids in database.
It's not blocking though, you can put a TODO if you prefer.

alexgarel (changes requested):
@aadarsh-ram kudos for your code which is correct and well organized.

I commented on some improvement that I see necessary (some are more on style).

I think we should really avoid short names like `a`, `b` etc. in our queries and use descriptive name, this really helps understand the code in a fluid way. (coming from outside, I see this more clearly as I review the code).

backend/editor/entries.py
```diff
@@ -21,6 +21,91 @@ def shutdown_db():
...
+def get_label(id):
+    """
+    Helper function for getting the label for a given id
+    """
+    if (id.startswith('stopword')): return 'STOPWORDS'
+    elif (id.startswith('synonym')): return 'SYNONYMS'
+    elif (id.startswith('__header__') or id.startswith('__footer__')): return 'TEXT'
+    else: return 'ENTRY'
+
+def create_node(label, entry, mainLanguageCode):
```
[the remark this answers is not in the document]
aadarsh-ram: Added a check in api.py

backend/editor/entries.py
aadarsh-ram: I separated the check for non-existing nodes, from the "creation of new relationships" part. Currently, the node creates a new code if it doesn't exist. After all nodes have been created, the relationships are built. Do you think I'd need to change anything here?

backend/editor/api.py
```diff
@@ -188,6 +189,25 @@ async def findFooter(response: Response):
...
+    """
+    Creating a new node in a taxonomy
+    """
+    incomingData = await request.json()
+    id = incomingData["id"]
+    main_language = incomingData["main_language"]
+    if (id == None):
+        raise ValueError("Invalid id: %s", id)
+    if (main_language == None):
+        raise ValueError("Invalid main language: %s", main_language)
```
alexgarel, line 203: You should raise [HTTPError](https://fastapi.tiangolo.com/tutorial/handling-errors/#use-httpexception) with [400 code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses)

backend/editor/api.py
```diff
@@ -242,4 +274,15 @@ async def editFooter(incomingData: Footer):
...
     result = update_nodes("TEXT", "__footer__", convertedData)
     updatedFooter = list(result)
-    return updatedFooter
\ No newline at end of file
+    return updatedFooter
+
+# Delete methods
+
+@app.delete("/nodes")
+async def deleteNode(request: Request):
```
alexgarel, line 282: BTW why didn't you choose to have:
```suggestion
@app.delete("/nodes/{id}")
async def deleteNode(request: Request):
```

backend/editor/entries.py
```diff
@@ -21,6 +21,98 @@ def shutdown_db():
...
+    end_node_label = get_label(end_node['id']) # Get current last node ID
+
+    # Rebuild relationships by inserting incoming node at the end
+    query = []
+    query.append(f"""MATCH (new_node:{label}) WHERE new_node.id = $id\n""")
+    query.append(f"""MATCH (last_node:{end_node_label}) WHERE last_node.id = $endnodeid\n""")
+    query.append(f"""MATCH (footer:TEXT) WHERE footer.id = "__footer__"\n""")
+    query.append(f"""CREATE (last_node)-[:is_before]->(new_node)\n""")
+    query.append(f"""CREATE (new_node)-[:is_before]->(footer)""")
+    result = session.run(" ".join(query), {"id": entry, "endnodeid": end_node['id']})
```
alexgarel, line 76:
```suggestion
    query = f"""
        MATCH (new_node:{label}) WHERE new_node.id = $id
        MATCH (last_node:{end_node_label}) WHERE last_node.id = $endnodeid
        MATCH (footer:TEXT) WHERE footer.id = "__footer__"
        CREATE (last_node)-[:is_before]->(new_node)
        CREATE (new_node)-[:is_before]->(footer)
    """
    result = session.run(query, {"id": entry, "endnodeid": end_node['id']})
```

backend/editor/entries.py
```diff
@@ -21,6 +21,98 @@ def shutdown_db():
...
+    start_node_label = get_label(start_node['id']) # Get current first node ID
+
+    # Rebuild relationships by inserting incoming node at the beginning
+    query = []
+    query.append(f"""MATCH (new_node:{label}) WHERE new_node.id = $id\n""")
+    query.append(f"""MATCH (first_node:{start_node_label}) WHERE first_node.id = $startnodeid\n""")
+    query.append(f"""MATCH (header:TEXT) WHERE header.id = "__header__"\n""")
+    query.append(f"""CREATE (new_node)-[:is_before]->(first_node)\n""")
+    query.append(f"""CREATE (header)-[:is_before]->(new_node)""")
+    result = session.run(" ".join(query), {"id": entry, "startnodeid": start_node['id']})
```
alexgarel, line 98:
```suggestion
    query= f"""
        MATCH (new_node:{label}) WHERE new_node.id = $id
        MATCH (first_node:{start_node_label}) WHERE first_node.id = $startnodeid
        MATCH (header:TEXT) WHERE header.id = "__header__"
        CREATE (new_node)-[:is_before]->(first_node)
        CREATE (header)-[:is_before]->(new_node)
    """
    result = session.run(query, {"id": entry, "startnodeid": start_node['id']})
```

backend/editor/entries.py
```diff
@@ -21,6 +21,98 @@ def shutdown_db():
...
+    # Finding node to be deleted using node ID
+    query = []
+    query.append(f"""MATCH (deleted_node:{label})-[:is_before]->(next_node) WHERE deleted_node.id = $id\n""")
+    query.append(f"""MATCH (previous_node)-[:is_before]->(deleted_node)\n""")
+    query.append(f"""DETACH DELETE (deleted_node)""")
+
+    # Rebuild relationships after deletion
+    query.append(f"""CREATE (previous_node)-[:is_before]->(next_node)""") 
+
+    result = session.run(" ".join(query), {"id": entry})
```
alexgarel, line 113:
```suggestion
    query = f"""
        // Find node to be deleted using node ID
        MATCH (deleted_node:{label})-[:is_before]->(next_node) WHERE deleted_node.id = $id
        MATCH (previous_node)-[:is_before]->(deleted_node)
        // Remove
        DETACH DELETE (deleted_node)
        // Rebuild relationships after deletion
        CREATE (previous_node)-[:is_before]->(next_node)
    """ 
    result = session.run(query, {"id": entry})
```

backend/editor/entries.py
```diff
@@ -94,11 +186,52 @@ def update_nodes(label, entry, incomingData):
...
+
+    # Delete relationships
+    for child in deleted_children:
+        query = []
+        query.append(f""" 
+            MATCH (deleted_child:ENTRY)-[rel:is_child_of]->(parent:ENTRY) 
+            WHERE parent.id = $id AND deleted_child.id = $child\n 
+        """)
+        query.append(f""" DELETE rel """)
+        session.run(" ".join(query), {"id": entry, "child": child})
```
alexgarel, line 215:
```suggestion
        query= f""" 
            MATCH (deleted_child:ENTRY)-[rel:is_child_of]->(parent:ENTRY) 
            WHERE parent.id = $id AND deleted_child.id = $child
            DELETE rel
        """
        session.run(query, {"id": entry, "child": child})
```
:-)

backend/editor/entries.py
```diff
@@ -94,11 +186,52 @@ def update_nodes(label, entry, incomingData):
...
+
+    # Check existence of node in new children
+    for child in added_children:
+        is_old_node = list(get_nodes("ENTRY", child))
+
+        # Create node if not exists
+        if (not is_old_node):
+            create_node("ENTRY", child, main_language_code)
+            add_node_to_end("ENTRY", child)
```
alexgarel, line 224:
```suggestion
            # TODO we would prefer to add the node just after the entry it is child of
            add_node_to_end("ENTRY", child)
```

backend/editor/entries.py
alexgarel: We may keep it like it, my concern was about performance:

You could do:
```python
    # create non existing node
    existing = session.run("""MATCH (child:ENTRY) WHERE id in $ids RETURN n.id""", ids=added_children)
    to_create = added_children - set(existing)
    for child in to_create:
        main_language_code = child.split(":", 1)[0]
        create_node("ENTRY", child, main_language_code)
    # create relations
    for child in added_children:
    ...
```

backend/editor/entries.py
```diff
@@ -94,11 +186,52 @@ def update_nodes(label, entry, incomingData):
...
+        query.append(f""" DELETE rel """)
+        session.run(" ".join(query), {"id": entry, "child": child})
+
+    # Check existence of node in new children
+    for child in added_children:
+        is_old_node = list(get_nodes("ENTRY", child))
+
+        # Create node if not exists
+        if (not is_old_node):
+            create_node("ENTRY", child, main_language_code)
```
alexgarel, line 223: main_language_code is not defined, is it ?

backend/editor/entries.py
```diff
@@ -94,11 +186,52 @@ def update_nodes(label, entry, incomingData):
...
+    # Stores result of last query executed
+    result = []
+    for child in added_children:
+        query = []
+        main_language_code = child.split(":", 1)[0]
+
+        # Create new relationships if it doesn't exist
+        query.append(f""" MATCH (parent:ENTRY), (new_child:ENTRY) WHERE parent.id = $id AND new_child.id = $child\n """)
+        query.append(f""" MERGE (new_child)-[r:is_child_of]->(parent) """)
+        result = session.run(" ".join(query), {"id": entry, "child": child})
```
alexgarel, line 235:
```suggestion
        main_language_code = child.split(":", 1)[0]
        # Create new relationships if it doesn't exist
        query = f"""
            MATCH (parent:ENTRY), (new_child:ENTRY) WHERE parent.id = $id AND new_child.id = $child
            MERGE (new_child)-[r:is_child_of]->(parent)
        """
        result = session.run(query, {"id": entry, "child": child})
```

alexgarel (commented):
Some suggestions to apply ! And there we are.

backend/editor/api.py
aadarsh-ram: I wanted to keep the `/nodes` endpoint uniform. So, GET, POST, DELETE perform different functions and the request body only must be changed.

alexgarel: Yes but delete is more like editEntry than list of nodes.

In REST when you use DELETE, it's on a url representing an entry.

aadarsh-ram: But, using `/entry/{entry}` would mean the deletion of an entry only. This endpoint would be able to delete any node in the taxonomy. So, I thought it should have the `/nodes` endpoint.

alexgarel (approved):
Thanks for the good work.

PR #86, opened 2022-09-16
build: Change parser workflow

aadarsh-ram:
### What
- Parser checks workflow runs for every PR
- It must be run only when any files inside the `parser` directory is changed.


alexgarel (approved):
LGTM, thanks

Issue #72, opened 2022-09-06
Add search capability (UI and backend)


alexgarel:
- [x] add full index to the graph when we create it (in parser)
- [x] add api point to make a full text search
- [x] create ui to launch a  search
- [x] create ui to display search results
- [ ] infinity scroll/pagination

PR #91, opened 2022-09-23
feat: Search functionality - Backend

aadarsh-ram:
### What
- Search functionality in backend

### Related Issue(s)
- #72


backend/editor/entries.py
```diff
@@ -242,3 +242,38 @@ def update_node_children(entry, new_children_ids):
...
+    # Fuzzy search has more priority, since it matches more strings
+    query = f"""
+        CALL db.index.fulltext.queryNodes("nodeSearchIds", $textqueryfuzzy) YIELD node, score
+        WHERE score > 0.2
+        RETURN node.id
+        UNION
+        CALL db.index.fulltext.queryNodes("nodeSearchTags", $textqueryfuzzy) YIELD node, score
+        WHERE score > 0.2
+        RETURN node.id
+    """
```
alexgarel, line 264: You can do all your queries in one go by introducing an "order by score" and multiplying the score by a constant (here I choose 5, but we make tweak more finely):


```suggestion
    query = f"""
        CALL db.index.fulltext.queryNodes("nodeSearchIds", $textqueryfuzzy)
        YIELD node, score as score_
        WHERE score_ > 0.2
        RETURN node.id, score_ * 5 as score
        ORDER BY score DESC
        UNION
        CALL db.index.fulltext.queryNodes("nodeSearchTags", $textqueryfuzzy)
        YIELD node, score as score_
        WHERE score_ > 0.2
        RETURN node.id, score_ * 5 as score
        ORDER BY score DESC
        CALL db.index.fulltext.queryNodes("nodeSearchIds", $textqueryexact) YIELD node, score
        WHERE score > 0.2
        RETURN node.id, score
        ORDER BY score DESC
        UNION
        CALL db.index.fulltext.queryNodes("nodeSearchTags", $textqueryexact) YIELD node, score
        WHERE score > 0.2
        RETURN node.id, score
        ORDER BY score DESC
    """

```

backend/editor/entries.py
```diff
@@ -242,3 +242,38 @@ def update_node_children(entry, new_children_ids):
...
     
     return result
+
+def full_text_search(text):
+    """
+    Helper function used for searching a taxonomy
+    """
+    normalized_text = re.sub(r"([^A-Za-z0-9])", r"\\\1", text) # Escape special characters
+    text_query_exact = "*" + normalized_text + '*'
+    text_query_fuzzy = normalized_text + "~"
```
alexgarel, line 252: I think we should do a special version for "nodeSearchIds" to instead use the id normalization function.
So that if I search "Strawberry yogurt", it will search "stawberry-yogurt~" and "*stawberry-yogurt*"

alexgarel (changes requested):
We are on a good way !
Great work.

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -1,7 +1,9 @@
 import logging
 import re
+import sys
 import unicodedata
 
+import iso639
```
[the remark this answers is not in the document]
alexgarel: Note that we could also harvest language code while parsing.

backend/editor/entries.py
alexgarel: a bit more complicated than I though !
Here is the real query after some test :-)

```suggestion
    query = """
        CALL {
                CALL db.index.fulltext.queryNodes("nodeSearchIds", $textqueryfuzzy)
                yield node, score as score_
                return node, score_ * 5 as score
            UNION
                CALL db.index.fulltext.queryNodes("nodeSearchTags",  $textqueryfuzzy)
                yield node, score as score_
                return node, score_ * 5 as score
            UNION
                CALL db.index.fulltext.queryNodes("nodeSearchIds",  $text_query_exact)
                yield node, score as score_
                return node, score_ as score
            UNION
                CALL db.index.fulltext.queryNodes("nodeSearchTags", $text_query_exact)
                yield node, score as score_
                return node, score_ as score 
        }
        with node.id as node, score

        RETURN node, sum(score) as score
        
        ORDER BY score DESC
    """

```

alexgarel: And a snippet to test it in neo4j:
```
CALL {
            CALL db.index.fulltext.queryNodes("nodeSearchIds", "yogurt~")
            yield node, score as score_
            return node, score_ * 5 as score
            UNION
            CALL db.index.fulltext.queryNodes("nodeSearchTags",  "yogurt~")
            yield node, score as score_
            return node, score_ * 5 as score
            UNION
            CALL db.index.fulltext.queryNodes("nodeSearchIds",  "*yog*")
            yield node, score as score_
            return node, score_ as score
            UNION
            CALL db.index.fulltext.queryNodes("nodeSearchTags", "*yog*")
            yield node, score as score_
            return node, score_ as score 
        }
        with node.id as node, score

        RETURN node, sum(score) as score
        
        ORDER BY score DESC
```

aadarsh-ram: @alexgarel this doesn't work for the query string "meat". The score is skewed towards "en:fake-meat", but "en:meat" is the closest string.

parser/openfoodfacts_taxonomy_parser/parser.py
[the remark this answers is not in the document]
aadarsh-ram: Yeah, but this way during creation of index, all possible language codes have been indexed. If new tags of different languages are added, there shouldn't be an issue.

backend/editor/entries.py
aadarsh-ram: @alexgarel I think we need to give a little less priority to the `nodeSearchIds` index during fuzzy search, and use take the average score instead of the sum.
```
CALL {
            CALL db.index.fulltext.queryNodes("nodeSearchIds", "yogurt~")
            yield node, score as score_
            return node, score_ * 3 as score
            UNION
            CALL db.index.fulltext.queryNodes("nodeSearchTags",  "yogurt~")
            yield node, score as score_
            return node, score_ * 5  as score
            UNION
            CALL db.index.fulltext.queryNodes("nodeSearchIds",  "*yogurt*")
            yield node, score as score_
            return node, score_ as score
            UNION
            CALL db.index.fulltext.queryNodes("nodeSearchTags", "*yogurt*")
            yield node, score as score_
            return node, score_ as score 
        }
        with node.id as node, score

        RETURN node, sum(score)/count(score) as score
        
        ORDER BY score DESC
```
Both query strings "yogurt" and "meat" worked perfectly with this. Is this okay?

backend/editor/entries.py
aadarsh-ram: Where exactly would this query come? How much priority would we need to give for this?

parser/openfoodfacts_taxonomy_parser/parser.py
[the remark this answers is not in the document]
alexgarel: You are completely right :-)

backend/editor/entries.py
alexgarel: Good to have tested more thoroughly.
your proposal this is perfectly fine to me.

backend/editor/entries.py
alexgarel: I just propose you to use a normalized version of text (using the same function as when you transform tag tot tag id) and build a text_id_query_exact and text_id_query_fuzzy and use as the parameter for the queryNodes on nodeSearchIds instead of text_query_exact and text_query_fuzzy

alexgarel: It's because I suspect current version not to work on tags_id with multi word searches. (like "Straweberry yogurt")

backend/editor/entries.py
```diff
@@ -247,33 +248,45 @@ def full_text_search(text):
     """
     Helper function used for searching a taxonomy
     """
-    normalized_text = re.sub(r"([^A-Za-z0-9])", r"\\\1", text) # Escape special characters
+    # Escape special characters
+    normalized_text = re.sub(r"([^A-Za-z0-9 _])", r"\\\1", text)
+    normalized_id_text = re.sub(r"([^A-Za-z0-9 _])", r"\\\1", normalizing(text))
```
alexgarel, line 253: I don't think you need to re-normalize it ? I mean, it's fine here to match on "-"

backend/editor/entries.py
```diff
@@ -247,33 +248,45 @@ def full_text_search(text):
...
+    # Escape special characters
+    normalized_text = re.sub(r"([^A-Za-z0-9 _])", r"\\\1", text)
+    normalized_id_text = re.sub(r"([^A-Za-z0-9 _])", r"\\\1", normalizing(text))
+
     text_query_exact = "*" + normalized_text + '*'
     text_query_fuzzy = normalized_text + "~"
-
-    # Fuzzy search on two indexes
+    text_id_query_exact = normalized_id_text + "~"
+    text_id_query_fuzzy = "*" + normalized_id_text + "*"
```
alexgarel, line 258: you invert the names ?

backend/editor/entries.py
```diff
@@ -247,33 +248,45 @@ def full_text_search(text):
...
+                CALL db.index.fulltext.queryNodes("nodeSearchIds",  $text_id_query_exact)
+                yield node, score as score_
+                return node, score_ as score
+            UNION
+                CALL db.index.fulltext.queryNodes("nodeSearchTags", $text_query_exact)
+                yield node, score as score_
+                return node, score_ as score 
+        }
+        with node.id as node, score
+        RETURN node, avg(score) as score
```
alexgarel, line 287: It's very strange to use an average instead of a sum here. 
It means if it matches both fuzzy and exact it may have a lower score ?

Eventually you want to use "max" ?

backend/editor/normalizer.py
```diff
@@ -0,0 +1,30 @@
+import re
```
alexgarel, line 1: It's ok for now, but as you move to integrate the parser, we will use it as a library, and we'd better have this function there, shared with parsing code.

alexgarel (changes requested):
Still some comments.

backend/editor/entries.py
aadarsh-ram: @alexgarel I'm not sure I'd want to use "max", since * searches can have more score over fuzzy searches. But, fuzzy searches would have more accuracy in terms of finding the closest match. So, an average of all the scores must be taken to figure out the best possible match.

aadarsh-ram: Atleast in my tests, "max" might give the wrong closest node, and "avg" will find the best possible match. @alexgarel how do you think I should proceed?

backend/editor/entries.py
```diff
@@ -250,7 +250,7 @@ def full_text_search(text):
     """
     # Escape special characters
     normalized_text = re.sub(r"([^A-Za-z0-9 _])", r"\\\1", text)
-    normalized_id_text = re.sub(r"([^A-Za-z0-9 _])", r"\\\1", normalizing(text))
+    normalized_id_text = re.sub(r"(-)", r"\\\1", normalizing(text))
```
alexgarel, line 253: Do you really think you have to normalize "-" to "\-" for the search to work ?

alexgarel (approved):
LGTM

backend/editor/entries.py
alexgarel: @aadarsh-ram could you look into this before merging (sorry this was a pending comment I forgot about.

aadarsh-ram: @alexgarel I'll look into it

alexgarel: @aadarsh-ram I will say just add a .replace("-", " ") after normalizing.

alexgarel (approved):
Great @aadarsh-ram :tada:

PR #93, opened 2022-09-27
fix: Using Neo4J transactions

### LINKED ISSUE #78: Link sessions to requests and handle transactions
Right now the initialization of the neo4j session is very rudimentary in the backend.

It is shared among requests and we don't handle transactions.

Having a concurrent application we really need to handle both of that, or this may lead to very inconsistent database states.

The simplest strategy is to have one session and transaction per request. The transaction should be rolled back if an exception pops up, and be committed otherwise (this is the default behavior of the transaction context manager provided by neo4j)

We can do that through a [middleware](https://fastapi.tiangolo.com/tutorial/middleware/) and open [session and transaction](https://neo4j.com/docs/api/python-driver/current/api.html#sessions-transactions) there.

The middleware will use the transaction context manager and call next_call inside it.
It has to store the returned transaction in a [contextvar](https://docs.python.org/3/library/contextvars.html) which is a global variable in the entries.py module.

Instead of using the session in entries.py (and everywhere else), we should instead get the transaction through the contextvar .

aadarsh-ram:
### What
- Using transactions for requests in order to handle them concurrently.

### Related bug(s)
- Fixes #78


backend/editor/api.py
```diff
@@ -41,16 +44,26 @@
...
-    shutdown_db()
+    driver.close()
+
+@app.middleware("http")
+async def transaction_creation(request: Request, call_next):
+    with driver.session() as session:
+        global txn
+        txn = session.begin_transaction()
+        response = await call_next(request)
+        return response
```
alexgarel, line 66: Put it in it's own file (with initialize_db and shutdown_db. So that you don't have cross import problems.

Maybe graph_db.py

backend/editor/api.py
```diff
@@ -41,16 +44,26 @@
...
-    Shutdown database
+    Close driver of Neo4J database
     """
-    shutdown_db()
+    driver.close()
+
+@app.middleware("http")
+async def transaction_creation(request: Request, call_next):
+    with driver.session() as session:
+        global txn
```
alexgarel, line 63: give at better name :-) (I know you are still experimenting) like current_transaction

backend/editor/api.py
```diff
@@ -82,7 +95,7 @@ async def findAllNodes(response: Response):
     """
     Get all nodes within taxonomy
     """
-    result = get_all_nodes("")
+    result = get_all_nodes("", txn)
```
alexgarel, line 98: Instead of passing txn around, if you put it in a graph_db.py file you can import it in entries.py and use it there. But beware as it is a global, you should not use:
```python
from .graph_db import current_transaction
...
current_transaction.do_something
```
Because you would miss updates.

You have to instead keep the namespace:

```python
from . import graph_db
...
graph_db.current_transaction.do_something
```

backend/editor/entries.py
```diff
@@ -2,24 +2,6 @@
...
-    uri = settings.uri
-    driver = GraphDatabase.driver(uri)
-    session = driver.session()
-
-def shutdown_db():
-    """
-    Close session and driver of Neo4J database
-    """
-    session.close()
-    driver.close()
```
alexgarel, line 22: As said above don't put it neither in entries nor in api, put it in it's own module.

backend/editor/api.py
alexgarel, line 66: You really have to use the context manager ! because here, you never end the transaction !
See https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.Transaction

```suggestion
        with session.begin_transaction() as _txn:
            # make it current transaction
            txn = _txn
            response = await call_next(request)
        txn = None
        return response
```

backend/editor/api.py
```diff
@@ -43,14 +43,24 @@ async def startup():
...
+    graph_db.shutdown_db()
+
+@app.middleware("http")
+async def initialize_neo4j_transactions(request: Request, call_next):
+    with graph_db.driver.session() as session:
+        with session.begin_transaction() as _txn:
+            # Make it the current transaction
+            graph_db.txn = _txn
+            response = await call_next(request)
+        graph_db.txn = None
```
alexgarel, line 62: If you're interested we can make this part a context manager in graph_db (so that all code managing db is in there):

```python
@contextlib.contextmanager
def TransactionCtx():
    """Transaction context will set global transaction txn for the code in context

    transaction is automatically rollback if an exception occurs within the context
    """
    global txn
    with driver.session() as session
        with session.begin_transaction() as _txn:
            txn = _txn
            yield txn
    txn = None
```

and here:

```python
@app.middleware("http")
async def initialize_neo4j_transactions(request: Request, call_next):
    with graph_db.TransactionCtx():
        response = await call_next(request)
```

backend/editor/graph_db.py
```diff
@@ -0,0 +1,18 @@
+from neo4j import GraphDatabase         # Interface with Neo4J
+from . import settings                  # Neo4J settings
+
+txn = None
```
alexgarel, line 4: Ok, now the next step (because we want to support concurrency) is to make it compatible with concurrency (in an async context).

For this, you will need a [context variable](https://docs.python.org/3/library/contextvars.html)

It's better to encapsulate "getting the transaction" in a function. So that in entries.py you have:
`transaction().run(....)`
(note that since you will use a function, you can now import the function directly: `from graph_bd import transaction()`)
This function should return current value of global context variable, or raise if it's not set (or set to None)

alexgarel (changes requested):
Cool, now let's introduce context variables and it will be ready.

alexgarel (approved):
Great !
Almost there, a small change required. However I approve so that you can merge right after that.

aadarsh-ram:
@alexgarel I think we can merge this after the search functionality PR is merged.

aadarsh-ram:
Thanks for your help @alexgarel! 👍

PR #94, opened 2022-09-30
docs: added contributing.md

alexgarel:
### What
Added Contributed.md by adapting the one from robotoff.


aadarsh-ram (approved):
@alexgarel LGTM. If required, you can add `CONTRIBUTING.md` as a top level file in guides, so that it can serve as a summary for any other guides in the future.

alexgarel:
>  If required, you can add CONTRIBUTING.md as a top level file in guides, so that it can serve as a summary for any other guides in the future.

Yes @aadarsh-ram I think it would be useful indeed, but it was in a hurry for hacktoberfest, let's make it more rational in the future :-)

PR #95, opened 2022-10-04
fix: UUIDs for tags of a node

aadarsh-ram:
### What
- The frontend requires a unique key for displaying each tag of a stopword or entry
- This key has to be generated/stored in the backend to maintain consistent UUIDs for each tag


aadarsh-ram:
Closing this in favor of #100

PR #97, opened 2022-10-05
fix: Change session to transactions

aadarsh-ram:
### What
- Used session.run() instead of get_current_transaction().run()


alexgarel (approved):
:+1:

PR #100, opened 2022-10-05
fix: Add UUIDs after fetching node

aadarsh-ram:
### What
- The frontend requires a unique key for displaying each tag of a stopword or entry
- The frontend generates a unique key for each node object after fetch

### DIFF
--- a/taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
+++ b/taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
@@ -7,6 +7,7 @@ import ListTranslations from "./ListTranslations";
 import ListAllEntryProperties from "./ListAllEntryProperties";
 import ListAllNonEntryInfo from "./ListAllNonEntryInfo";
 import { createURL, getIdType } from "./createURL";
+import * as uuid from "uuid";
 
 /**
  * Component used for rendering node information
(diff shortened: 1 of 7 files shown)


taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -27,7 +28,23 @@ const AccumulateAllComponents = ({ id }) => {
...
+                    duplicateNode[key+'_uuid'] = [];
+                    duplicateNode[key].forEach(() => {
+                        duplicateNode[key+'_uuid'].push(uuid.v4());
+                    })
+                }
+                else if (key.startsWith('prop')) {
+                    duplicateNode[key+'_uuid'] = [uuid.v4()];
+                }
+            })
+            console.log(duplicateNode)
```
nobeeakon, line 45: nit: please remove the console.log

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -27,7 +28,23 @@ const AccumulateAllComponents = ({ id }) => {
 
     // Setting state of node after fetch
     useEffect(() => {
-        setNodeObject(node?.[0]);
+        if (node) {
```
nobeeakon, line 31: should it do something if `!not` ? like setting to `null` or something?

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
```diff
@@ -47,15 +64,22 @@ const AccumulateAllComponents = ({ id }) => {
     const handleSubmit = () => {
         if (!nodeObject) return
         const {id, ...data} = nodeObject // ID not allowed in POST
-        let allUrlsAndData = [[url, data]]
+        const dataToBeSent = {};
```
nobeeakon, line 67: nice naming

nobeeakon (changes requested):
nice work

taxonomy-editor-frontend/src/pages/editentry/AccumulateAllComponents.jsx
aadarsh-ram: I don't think it's necessary since the initial state is already null. Only when the node info is populated, we need to update the state.

nobeeakon: what about if url changes and data is renewed?

nobeeakon:
great work !!

nobeeakon (approved):

PR #101, opened 2022-10-06
fix: Add multiple labels in backend

### LINKED ISSUE #69: Change entries.py to support branch and taxonomy name
We want to edit a taxonomy with a branch name and a taxonomy name.

Change entries.py so that it defines a class (I propose TaxonomyGraph).

The class takes branch (branch name) and taxonomy name as parameters.

Current functions become methods of this class.

So that the caller can do:
```python
taxonomy = TaxonomyGraph(branch_name, taxonomy_name)
taxonomy.get_nodes(...)
```

Modify those methods to add the branch_name and taxonomy_name labels. First is prefixed by `b_`, second is prefixed by  `t_`

aadarsh-ram:
### What
- Add multiple labels (to support branch and taxonomy labelling) in parser
- Add multiple labels in unparser and tests
- Changed API endpoints accordingly to support different branches and taxonomies
- Wrapped entries.py in a class for easier use
- Added import functionality
- Added export functionality

### Related issue(s)
- Fixes #69


parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -87,35 +96,35 @@ def file_iter(self, filename, start=0):
...
             line,
         )
 
-        # removing excess "-"
-        line = re.sub(r"-+", "-", line)
-        line = line.strip("-")
+        # Removing excess "-"
+        line = re.sub(r"-+", char, line)
+        line = line.strip(char)
         return line
```
alexgarel, line 124: I would like to avoid code duplication.

Backend must be able to use parser has a library (you can (add `-e ../parser/openfoodfacts_taxonomy_parser` in requirements.txt, and in the Docker we have to copy the parser path in the docker)).

So:
1. here we should put normalizing in a separate function (in normalizer.py is even better)
2. call this function in this method
3. import from parser and call it in backend instead of having a copy of normalizer.py there.

It's ok if you want to do it on next step (creating a project), but this really has to be done. Code duplication means possible divergence, which is bad !

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -418,27 +428,32 @@ def delete_used_properties(self):
...
-            CREATE FULLTEXT INDEX nodeSearchIds FOR (n:ENTRY) ON EACH [n.id]
-            OPTIONS {indexConfig: {`fulltext.analyzer`: 'keyword'}}
-        """
-        self.session.run(query)
+    def create_fulltext_index(self, multi_label, branch_name):
+        """Create indexes for search"""
+        index_name = multi_label.replace(":", "_")
+        query = [
+            f"""CREATE FULLTEXT INDEX {index_name+'_SearchIds'}
+            FOR (n:{'b_'+branch_name}) ON EACH [n.id]\n"""
```
alexgarel, line 436: Why not use multilabel here ?

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -418,27 +428,32 @@ def delete_used_properties(self):
...
+        self.session.run("".join(query))
 
         language_codes = [lang.alpha2 for lang in list(iso639.languages) if lang.alpha2 != ""]
         tags_prefixed_lc = ["n.tags_" + lc + "_str" for lc in language_codes]
         tags_prefixed_lc = ", ".join(tags_prefixed_lc)
-        query = (
-            f"""CREATE FULLTEXT INDEX nodeSearchTags FOR (n:ENTRY) ON EACH [{tags_prefixed_lc}]"""
-        )
+        query = f"""CREATE FULLTEXT INDEX {index_name+'_SearchTags'}
+            FOR (n:{'b_'+branch_name}) ON EACH [{tags_prefixed_lc}]"""
```
alexgarel, line 445: Why not use multilabel here ?

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -418,27 +428,32 @@ def delete_used_properties(self):
...
-    def __call__(self, filename):
-        """process the file"""
-        self.create_nodes(filename)
-        self.create_child_link()
-        self.create_previous_link()
-        self.create_fulltext_index()
+    def __call__(self, filename, branch_name):
+        """Process the file"""
+        filename = self.normalized_filename(filename)
+        branch_name = self.normalizing(branch_name, char="_")
```
alexgarel, line 451: It's pefectly ok to use "-" for branch name, indeed I think it's more readable.
I name my branches: taxonomy-feat-do-blah.

parser/openfoodfacts_taxonomy_parser/unparser.py
```diff
@@ -118,14 +127,16 @@ def rewrite_file(self, filename, lines):
...
 
 
 if __name__ == "__main__":
-    import sys
-
     filename = sys.argv[1] if len(sys.argv) > 1 else "test"
+    branch_name = sys.argv[2] if len(sys.argv) > 1 else "branch"
     write = WriteTaxonomy()
-    write(filename)
+    write(filename, branch_name)
```
alexgarel, line 142: as for Parser, make taxonomy name a distinct parameter, not necessarily correlated to filename.

parser/tests/integration/test_parse_unparse_integration.py
```diff
@@ -6,12 +6,13 @@
...
 # taxonomy in text format : test.txt
 TEST_TAXONOMY_TXT = str(pathlib.Path(__file__).parent.parent / "data" / "test.txt")
+MULTI_LABEL = "t_test:b_branch"
 
 
 @pytest.fixture(autouse=True)
 def test_setup(neo4j):
     # delete all the nodes and relations in the database
-    query = "MATCH (n) DETACH DELETE n"
+    query = f"MATCH (n:{MULTI_LABEL}) DETACH DELETE n"
```
alexgarel, line 15: (not a big deal) You could have hardcoded it, it maybe easier to read.

parser/tests/integration/test_parse_unparse_integration.py
```diff
@@ -21,17 +22,17 @@ def test_round_trip():
...
     assert number_of_nodes == 13
     session.close()
 
     # dump taxonomy back
     test_dumper = unparser.WriteTaxonomy()
-    lines = list(test_dumper.iter_lines())
+    lines = list(test_dumper.iter_lines(MULTI_LABEL))
 
     original_lines = [line.rstrip("\n") for line in open(TEST_TAXONOMY_TXT)]
     # expected result is close to original file with a few tweaks
```
alexgarel, line 38: With current test, we are not testing that there is no interference between to parsing.

It would be important to add a quick test where we parse twice the same taxonomy with two different branch name, and verify it's all ok on both taxonomy.

backend/editor/entries.py
```diff
@@ -5,270 +5,284 @@
...
+        """
+        Helper function used for updation of node with given id and label
+        """
+        # Sanity check keys
+        for key in new_node_keys.keys():
+            if not re.match(r"^\w+$", key) or key == "id":
+                raise ValueError("Invalid key: %s", key)
+        
+        # Get current node information and deleted keys
+        curr_node = self.get_nodes(label, entry, self.multi_label).data()[0]['n']
```
alexgarel, line 166: I think we don't need the multi_label parameter:
```suggestion
        curr_node = self.get_nodes(label, entry).data()[0]['n']

```

backend/editor/entries.py
```diff
@@ -5,270 +5,284 @@
...
-    # Create non-existing nodes
-    query = """MATCH (child:ENTRY) WHERE child.id in $ids RETURN child.id"""
-    existing_ids = [record['child.id'] for record in get_current_transaction().run(query, ids=list(added_children))]
-    to_create = added_children - set(existing_ids)
+        for child in to_create:
+            main_language_code = child.split(":", 1)[0]
+            self.create_node("ENTRY", child, main_language_code, self.multi_label)
+            
+            # TODO: We would prefer to add the node just after its parent entry
+            self.add_node_to_end("ENTRY", child, self.multi_label)
```
alexgarel, line 222:
```suggestion
            self.add_node_to_end("ENTRY", child)

```

backend/editor/entries.py
```diff
@@ -5,270 +5,284 @@
...
+        existing_ids = [record['child.id'] for record in get_current_transaction().run(query, ids=list(added_children))]
+        to_create = added_children - set(existing_ids)
 
-    # Create non-existing nodes
-    query = """MATCH (child:ENTRY) WHERE child.id in $ids RETURN child.id"""
-    existing_ids = [record['child.id'] for record in get_current_transaction().run(query, ids=list(added_children))]
-    to_create = added_children - set(existing_ids)
+        for child in to_create:
+            main_language_code = child.split(":", 1)[0]
+            self.create_node("ENTRY", child, main_language_code, self.multi_label)
```
alexgarel, line 219:
```suggestion
            self.create_node("ENTRY", child, main_language_code)

```

parser/openfoodfacts_taxonomy_parser/parser.py
aadarsh-ram: It seems multiple labels are not allowed for indexes, hence I restricted the label to the branch name.

parser/openfoodfacts_taxonomy_parser/parser.py
aadarsh-ram: Ah yeah, but neo4j doesn't allow hyphens while creating indexes. So, I changed to a underscore delimiter.

parser/openfoodfacts_taxonomy_parser/parser.py
aadarsh-ram: @alexgarel I changed up all the imports and called the functions appropriately. I'm facing some issues with changing the docker related files. Would you be able to help me out?

parser/tests/integration/test_parse_unparse_integration.py
aadarsh-ram: @alexgarel I added a test (`test_two_branch_round_trip()`) for the same.

parser/openfoodfacts_taxonomy_parser/parser.py
alexgarel: hum that's really sad … I think thus we should change our strategy…

That is add three labels on entries at import time: branch_name, taxonomy_name and a blend of both.
The blend would be g_<branch_name>_<taxonomy_name> (prefix g like graph, but you may have a better idea !).

So throughout the code, you will use g_<branch_name>_<taxonomy_name>
But I would like to keep the b_ and t_ label, for later (when we would allow editing more than one taxonomy in the same project).

Are you ok to do this refactor ? Do you want to do it in a second phase (merge that and do it in a new PR ?)

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -1,13 +1,12 @@
...
 import re
 import sys
-import unicodedata
 
 import iso639
-import unidecode
 from neo4j import GraphDatabase
 
 from .exception import DuplicateIDError
+from .normalizer import normalizing
```
alexgarel, line 9: Thank you for doing the refactor :-)

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -447,5 +432,7 @@ def __call__(self, filename):
         handlers=[logging.FileHandler(filename="parser.log", encoding="utf-8")], level=logging.INFO
     )
     filename = sys.argv[1] if len(sys.argv) > 1 else "test"
+    branch_name = sys.argv[2] if len(sys.argv) > 1 else "branch"
+    taxonomy_name = sys.argv[3] if len(sys.argv) > 1 else "test"
```
alexgarel, line 436:
```suggestion
    taxonomy_name = sys.argv[3] if len(sys.argv) > 1 else filename.rsplit(".", 1)[0]
```

alexgarel (approved):
LGTM but two small suggestions, and a possible refactor but that we can do an a separate PR (as you want).

Note @aadarsh-ram merging this PR will brake the interface (as it still use the old API…). Is it ok ?

parser/openfoodfacts_taxonomy_parser/parser.py
aadarsh-ram: I think I'll do this refactor over here, and I'll simultaneously also give the frontend multilabel PR (it's almost complete). So, when both PRs are ready, they can be merged together. Does this sound good?

backend/editor/entries.py
```diff
@@ -10,8 +10,8 @@ class TaxonomyGraph:
     """Class for database operations"""
     
     def __init__(self, branch_name, taxonomy_name):
-        self.multi_label = ('t_'+taxonomy_name)+':'+('b_'+branch_name)
-        self.multi_label_search = ('t_'+taxonomy_name)+'_'+('b_'+branch_name)
+        self.project_name = 'p_' + taxonomy_name + '_' + branch_name
+        self.multi_label = self.project_name + ':' + ("t_" + taxonomy_name) + ":" + ("b_" + branch_name)
```
alexgarel, line 14: you only need the project name here, as this is redundant with the other labels.

We keep the other label just for future work (or being able to explore the database independently).

alexgarel (approved):
Ok perfect :-)

aadarsh-ram:
@alexgarel There seems to be some issue with the docker build in this workflow: https://github.com/openfoodfacts/taxonomy-editor/actions/runs/3399253092/jobs/5652877236#step:3:330

This issue occurs resolving conflicts in the backend Dockerfile. So, please do take a look at it.

alexgarel (approved):
LGTM.

PR #102, opened 2022-10-07
fix: Add score > 0 condition for search

aadarsh-ram:
### What
- Neo4J gives a score of 0 for nodes that don't match sometimes
- To account for that, a "where" clause with "score > 0" must be added in the backend


alexgarel (approved):
:+1:

PR #104, opened 2022-10-09
feat: Add new endpoint for fetching root nodes

aadarsh-ram:
### What
- Fetches all root nodes in given taxonomy
- Frontend will render these root nodes for user's convenience


alexgarel (approved):
perfect.

aadarsh-ram:
@alexgarel I'll merge this and we can merge #101 after the frontend is ready.

PR #105, opened 2022-10-10
build: support websocket in dev setup

alexgarel:
In dev react tries to open a websocket to the node server (for auto-refresh, I think). This was not supported, now this is.


aadarsh-ram (approved):
I checked out this PR and the frontend works! Thanks for the help! :)

Issue #62, opened 2022-08-24
Implement how to find the api in production 


alexgarel:
In production, the react application is a bunch of static files.
But we have one fundamental parameter: where is the api located ?
When we have the api, we could ask it for more parameters if needed.

This is discussed in [create-react-app documentation on environment variables](https://create-react-app.dev/docs/adding-custom-environment-variables).
Given solution are
1. serve a dynamic file for the initial html
2. rebuild for every deployment

I would like to avoid 2 as it adds complexity and less predictability.

Other solutions are:
1. The initial html could be served by the api (with a special route on nginx).
2. get the api url by simply changing current url prefix from ui to api

The solution 2. seems better to me, it's simple and flexible enough. But we have to deal with people using a local deployment, thus using the proxy of development server (not very complicated I think).

alexgarel:
@nobeeakon do you have any advice here ?

alexgarel:
blocking #73

PR #106, opened 2022-10-10
build: improve api host discovery

alexgarel:
fixes:
- #62


aadarsh-ram (approved):
@alexgarel I checked out both the PRs and everything works!

PR #110, opened 2022-10-10
fix: Change routes for including branch and taxonomy names

aadarsh-ram:
### What
- Change routes to account for branch name and taxonomy name


taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
```diff
@@ -69,14 +68,14 @@ const ListEntryChildren = ({url, setUpdateNodeChildren}) => {
         <Box>
             <Stack direction="row" alignItems="center">
                 <Typography sx={{ml: 4}} variant='h5'>Children</Typography>
-                <IconButton sx={{ml: 1, color: "#808080"}} onClick={handleOpen}>
+                <IconButton sx={{ml: 1, color: "#808080"}} onClick={handleOpenDialog}>
```
nobeeakon, line 71: please have this color as a `const`

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+    const navigate = useNavigate();
+
+    useEffect(() => { 
+        const renderedProjects = [];
+        if (incomingData) {
+            console.log(incomingData);
+            incomingData.forEach((element) => {
+                const projectNode = element[0];
+                renderedProjects.push({
+                    'id' : Math.random().toString(),
```
nobeeakon, line 23: nit: please add a note that this is used as react component key. May be confusing for some one looking at the code for the first time.

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+                    options={{
+                        actionsColumnIndex: -1, addRowPosition: "last",
+                        showTitle: false,
+                    }}
+                    actions={[
+                        {
+                            icon: () => <EditIcon />,
+                            tooltip: "Edit project",
+                            onClick: (event, rowData) => {
+                                navigate(`/${toSnakeCase(rowData['taxonomyName'])}/${rowData['branchName']}/entry`);
```
nobeeakon, line 67: is it possible to replace this with  a `<Link>` ?

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
```diff
@@ -0,0 +1,77 @@
...
+                    'branchName' : projectNode['branch_name']
+                })
+            })
+        }
+        setProjectData(renderedProjects);
+    }, [incomingData]);
+
+    // Check error in fetch
+    if (isError) {
+        return (<Typography variant='h5'>{errorMessage}</Typography>)
```
nobeeakon, line 35: for some reason I never put attention of this, but are the errors returned from the backed user readable? can we ensure that's the case for all of the errors? if not I think it is ok to just show: 'Ooops, something went wrong :(' Maybe have a component for that case or something.

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+            justifyContent="center"
+            >
+                <Typography sx={{mt: 4}} variant="h3">Start a project</Typography>
+                <Stack sx={{mt: 4, mb: 4}} direction="row" alignItems="center">
+                    <Typography sx={{mr: 4}} variant="h5">Taxonomy Name</Typography>
+                    <Autocomplete
+                        sx={{width: 265}}
+                        options={ TAXONOMY_NAMES }
+                        onChange={(e, selectedTaxonomy) => {
+                            if (selectedTaxonomy) setTaxonomyName(selectedTaxonomy.toLowerCase().replaceAll(/\s/g, '_'));
```
nobeeakon, line 51: is this a case for the functions you created for parsing the strings?

taxonomy-editor-frontend/src/constants.js
```diff
@@ -17,3 +17,33 @@ function taxonomyApiUrlFromUi(location) {
 }
 export const API_URL = taxonomyApiUrlFromUi(window.location);
 export const ENTER_KEYCODE = 13;
+export const TAXONOMY_NAMES = [
```
nobeeakon, line 20: please add a comment of where this taxonomies come from, to have an idea of when they need to be updated or where to check for that

taxonomy-editor-frontend/src/App.js
```diff
@@ -17,17 +20,20 @@ const theme = createTheme({
...
 
 function App() {
+  const [displayedPages, setDisplayedPages] = useState([]);
   return (
     <ThemeProvider theme={theme}>
     <CssBaseline />
     <Router>
-        <ResponsiveAppBar />
+        <ResponsiveAppBar displayedPages={displayedPages}/>
         <div className="App">
```
nobeeakon, line 29: please remove the className

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+const StartProject = () => {
+    const [branchName, setBranchName] = useState("")
+    const [taxonomyName, setTaxonomyName] = useState("additives")
+    const [description, setDescription] = useState("");
+    const [loading, setLoading] = useState(false); 
+    const [open, setOpen] = useState(false);
+    const [errorMessage, setErrorMessage] = useState("")
+    const navigate = useNavigate();
+    
+    function handleSubmit() {
```
nobeeakon, line 17: nit: let's try to stick with `const handleSubmit = () => `

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+            method : 'POST',
+            headers: {"Content-Type" : "application/json"},
+            body: JSON.stringify(dataToBeSent)
+        }).then((response) => {
+            if (!response.ok) {
+                return response.json();
+            }
+            navigate(`/${taxonomyName}/${branchName}/entry`)
+        }).then((responseBody) => {
+            setErrorMessage(responseBody.detail)
```
nobeeakon, line 31: are these errors user friendly ?

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+            setOpen(true); setLoading(false);
+        })
```
nobeeakon, line 33: `.catch(() => {})`?

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+    function handleClose() {setOpen(false);}
```
nobeeakon, line 35: nit: let's try to stick with `const handleSubmit = () =>`

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+import { useNavigate } from "react-router-dom";
+import { TAXONOMY_NAMES } from "../../constants";
+import { createBaseURL } from "../editentry/createURL";
+
+const StartProject = () => {
+    const [branchName, setBranchName] = useState("")
+    const [taxonomyName, setTaxonomyName] = useState("additives")
+    const [description, setDescription] = useState("");
+    const [loading, setLoading] = useState(false); 
+    const [open, setOpen] = useState(false);
```
nobeeakon, line 13: what about just open the thing when there's an error message, and when closing remove set error message to null? that way we can remove 1 state and make the logic cleaner?

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,100 @@
+import { Typography, Box, Grid, TextField, Stack, Autocomplete, Snackbar, Alert } from "@mui/material";
+import LoadingButton from '@mui/lab/LoadingButton';
```
nobeeakon, line 2: I guess if this is the only use for it, I'd prefer not to have an additional dependecy. Alternatively we could show a spinner in the button, or disable it while loading.

taxonomy-editor-frontend/package.json
```diff
@@ -7,6 +7,7 @@
     "@emotion/styled": "^11.9.3",
     "@material-table/core": "^0.2.35",
     "@mui/icons-material": "^5.8.4",
+    "@mui/lab": "^5.0.0-alpha.103",
```
nobeeakon, line 10: from what I could see the only thing we are using from it is a loading button, maybe we could show a loading spinner (we already have the icons package). Don't see it adds much value for the moment, and would be another dependecy

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+                        minRows={4}
+                        multiline
+                        onChange = {event => {
+                            setDescription(event.target.value)
+                        }}
+                        value={description} 
+                        variant="outlined" />
+                </Stack>
+                {/* Button for submitting edits */}
+                <LoadingButton
```
nobeeakon, line 82: I guess if this is the only use for it, I'd prefer not to have an additional dependecy. Alternatively we could show a spinner in the button, or disable it while loading.

taxonomy-editor-frontend/src/pages/allentries/index.jsx
```diff
@@ -38,7 +50,7 @@ const Entry = () => {
     function handleAddNode() {
```
nobeeakon, line 50: for these and others, let's try to stick to `const handleAddNode = () => {...}` we are mixing both synthax

taxonomy-editor-frontend/src/pages/allentries/index.jsx
```diff
@@ -29,6 +33,14 @@ const Entry = () => {
     const [openAddDialog, setOpenAddDialog] = useState(false);
     const [openSuccessSnackbar, setOpenSuccessSnackbar] = useState(false);
 
+    // Set url prefix for navbar component
+    useEffect(() => {
+        setDisplayedPages([
```
nobeeakon, line 38: thanks for the comment, really useful with `useEffect`. An alternative is using something like https://sergiodxa.com/articles/pro-tip-name-your-useeffect-functions  , it is not very common but I like it

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
aadarsh-ram: I don't think so, since it is dependent on the incoming `rowData`.

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
aadarsh-ram: I think it's currently just "Failed to fetch". I think we can add a component for that, or change it to "Oops. something went wrong :(".

taxonomy-editor-frontend/src/pages/startproject/index.jsx
aadarsh-ram: Yes, the errors returned by the backend are user friendly.

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
```diff
@@ -105,12 +104,10 @@ const ListEntryChildren = ({url, setUpdateNodeChildren}) => {
                         onKeyPress={(e) => { (e.keyCode === ENTER_KEYCODE) && handleAddChild(e) }} 
                         onChange={(e) => { 
                             setNewLanguageCode(e.target.value);
-                            const validateBool = ISO6391.validate(e.target.value);
-                            validateBool ? setisValidLC(true) : setisValidLC(false);
-                            validateBool ? setBtnDisabled(false) : setBtnDisabled(true);
+                            setIsValidLanguageCode(ISO6391.validate(e.target.value));
```
nobeeakon, line 107: could this state be replaced by a derived state? `const isValidLanguageCode = ISO...validate(newLanguageCode)`

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
```diff
@@ -13,13 +13,12 @@ import ISO6391 from 'iso-639-1';
...
-const ListEntryChildren = ({url, setUpdateNodeChildren}) => {
+const ListEntryChildren = ({url, urlPrefix, setUpdateNodeChildren}) => {
     const [relations, setRelations] = useState(null);
     const [newChild, setNewChild] = useState(null);
     const [newLanguageCode, setNewLanguageCode] = useState(null);
-    const [open, setOpen] = useState(false); // Used for Dialog component
-    const [btnDisabled, setBtnDisabled] = useState(true) // For enabling or disabling Dialog button
-    const [isValidLC, setisValidLC] = useState(false); // Used for validating a new LC
+    const [openDialog, setOpenDialog] = useState(false); // Used for Dialog component
+    const [isValidLanguageCode, setIsValidLanguageCode] = useState(false); // Used for validating a new LC
```
nobeeakon, line 21: could this be replaced by a derived state? makes it easier to maintain it in sync with the language code

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
```diff
@@ -36,8 +35,8 @@ const ListEntryChildren = ({url, setUpdateNodeChildren}) => {
     }, [incomingData, setUpdateNodeChildren])
 
     // Helper functions for Dialog component
-    function handleClose() { setOpen(false); }
-    function handleOpen() { setOpen(true); }
+    function handleCloseDialog() { setOpenDialog(false); }
+    function handleOpenDialog() { setOpenDialog(true); }
```
nobeeakon, line 39: we are mixing `function  ... ()` with `const ... = () => {}`. Let's stick with `const .. = () => {}`, just for consistency

taxonomy-editor-frontend/src/pages/editentry/index.jsx
```diff
@@ -1,27 +1,29 @@
...
-import { API_URL } from "../../constants";
+import { createBaseURL } from "./createURL";
 
-const EditEntry = () => {
-    const { id } = useParams();
-    const url = API_URL+'nodes';
+const EditEntry = ({setDisplayedPages}) => {
+    const { taxonomyName, branchName, id } = useParams();
+    const urlPrefix = `${taxonomyName}/${branchName}/`;
+    const url = createBaseURL(taxonomyName, branchName);
```
nobeeakon, line 16: nit: `baseURL`?

taxonomy-editor-frontend/src/pages/editentry/index.jsx
```diff
@@ -1,27 +1,29 @@
...
     const [openDeleteDialog, setOpenDeleteDialog] = useState(false);
     const [openSuccessDialog, setOpenSuccessDialog] = useState(false);
-    const navigate = useNavigate();
 
-    // Handler function for button clicks
-    const handleClick = (event) => {
-        event.preventDefault();
-        navigate('/entry');
-    }
+    // Set url prefix for navbar component
```
nobeeakon, line 20: nice, thanks for the comment

taxonomy-editor-frontend/src/pages/search/SearchResults.jsx
```diff
@@ -18,9 +17,12 @@ import DialogContent from '@mui/material/DialogContent';
 import DialogTitle from '@mui/material/DialogTitle';
 import Select from '@mui/material/Select';
 import ISO6391 from 'iso-639-1';
+import { createBaseURL } from "../editentry/createURL";
 
-const SearchResults = ({query}) => {
-    const { data: nodeIds, isPending, isError, isSuccess, errorMessage } = useFetch(`${API_URL}search?query=${encodeURI(query)}`);
+const SearchResults = ({query, taxonomyName, branchName}) => {
+    const url = createBaseURL(taxonomyName, branchName);
```
nobeeakon, line 23: nit: `baseUrl`? think at least it gives a hint that is comming from the `createBaseURL` function, and `url` can be a bit more cryptic

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
aadarsh-ram: I think `isValidLanguageCode` should be a state, since the Button component needs to be disabled/enabled based on it.

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
aadarsh-ram: Yep, added a constant called `greyHexCode`

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
aadarsh-ram: Added a comment for that

taxonomy-editor-frontend/src/pages/startproject/index.jsx
aadarsh-ram: Ah yeah, changed it to the function `toSnakeCase`

taxonomy-editor-frontend/src/constants.js
aadarsh-ram: I added a wiki doc for a list of taxonomies currently in Open Food Facts

taxonomy-editor-frontend/src/pages/startproject/index.jsx
aadarsh-ram: I refactored the code in such a way that the error message comes to the catch block.

taxonomy-editor-frontend/src/pages/startproject/index.jsx
aadarsh-ram: Yep, removed the `open` state and used a single state

taxonomy-editor-frontend/src/pages/startproject/index.jsx
aadarsh-ram: Yep, I created my own function `LoadingButton` to replace this

taxonomy-editor-frontend/package.json
aadarsh-ram: Ok, I removed the dependency.

taxonomy-editor-frontend/src/pages/allentries/index.jsx
aadarsh-ram: Yep, I refactored the code to use this syntax everywhere

taxonomy-editor-frontend/src/pages/allentries/index.jsx
aadarsh-ram: Cool, I added a name for the useEffect

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
aadarsh-ram: I think `isValidLanguageCode` should be a state, since the Button component needs to be disabled/enabled based on it.

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
nobeeakon: yes, think it would be ok to have it as derived and thus we can get rid of one state

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
nobeeakon: yes, think will be better to have it as derived state as it will change every time the `newLanguageCode` changes. And is easier to maintain in sync

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
```diff
@@ -13,13 +13,13 @@ import ISO6391 from 'iso-639-1';
...
+const ListEntryChildren = ({url, urlPrefix, setUpdateNodeChildren}) => {
     const [relations, setRelations] = useState(null);
     const [newChild, setNewChild] = useState(null);
     const [newLanguageCode, setNewLanguageCode] = useState(null);
-    const [open, setOpen] = useState(false); // Used for Dialog component
-    const [btnDisabled, setBtnDisabled] = useState(true) // For enabling or disabling Dialog button
-    const [isValidLC, setisValidLC] = useState(false); // Used for validating a new LC
+    const [openDialog, setOpenDialog] = useState(false); // Used for Dialog component
+    const [isValidLanguageCode, setIsValidLanguageCode] = useState(false); // Used for validating a new LC
+    const greyHexCode = "#808080";
```
nobeeakon, line 22: this should live in the `constants` file, not in here as this color is used several times. So you define it once and use it every time you need it. Defining it several times makes it difficult to ensure consistency

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
nobeeakon: Think we can leave it for now as is but, I'd prefer to replace it with something generic. Don't think it makes a lot of sense to let the user if the error was one of many possible fails (server side, internet connection, etc).

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+        setLoading(true);
+        const dataToBeSent = {'description' : description};
+        fetch(baseUrl+'import', {
+            method : 'POST',
+            headers: {"Content-Type" : "application/json"},
+            body: JSON.stringify(dataToBeSent)
+        }).then(async (response) => {
+            const responseBody = await response.json();
+            if (!response.ok) {
+                throw Error(responseBody.detail)
```
nobeeakon, line 27: guess the synthax should be `throw new Error()`. From a personal perspective, I think is more error prone to use `throw` inside promises/async, I honestly prefer `Promise.reject(new Error(...))` as I know it will returned as  a promise rejection, instead of a thrown error that will be hanled as a promise rejection (but in some cases is not)

nobeeakon: are we 100% sure that all responseBody that are not ok will have a `.detail` ? (like for example what happens if is an error that is not happening in the server, or that we are not sending, etc...) maybe use another check for that `if (!response.ok && responseBody.detail) {`

taxonomy-editor-frontend/src/pages/startproject/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+            }
+            navigate(`/${taxonomyName}/${branchName}/entry`)
+        }).catch((detail) => {
+            setErrorMessage(detail.message); setLoading(false);
```
nobeeakon, line 31: mmm ,  we need to check that the error has a `.message`. Also not 100% sure we want to show whatever is returned from the backend (like what happens with localization, and how can we ensure that this error has the information that is user friendly), but would need to see what errors are returned (like how different they are and if it is worth to handle it in this way). Guess for now I'd tend to think is safer not to have it like this and just return something simple as `unable to save`

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
aadarsh-ram: Yep, changed it

taxonomy-editor-frontend/src/pages/startproject/index.jsx
aadarsh-ram: It would be always having a detail, but I'll anyways add the check

taxonomy-editor-frontend/src/pages/startproject/index.jsx
aadarsh-ram: The API does return meaningful messages currently. But, it should be safer to keep a simple message.

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
aadarsh-ram: Yep, you're right. I made the changes.

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
aadarsh-ram: Yep, I made the changes.

taxonomy-editor-frontend/src/pages/startproject/index.jsx
nobeeakon: I think there's actually no way to be 100% sure about it, like if we control the response body then it will have it if we add it, but it can be `!ok` for reasons outisde of our control (server errors, network connections, etc) and in those circumstances it won't be possible to know if that will happen, which is the case for all errors (you can be sure of what you do/throw, but it can fail for whatever reason other than the cases you control)

taxonomy-editor-frontend/src/pages/startproject/index.jsx
nobeeakon: cool, yep, another issue is talked in the past if we want to have internationalization (translation) which will add the need to get a localized (translated) string from the API. I think is just fine to show a generic message.

nobeeakon (approved):
nice work!!!

aadarsh-ram:
Thanks @nobeeakon for all the help! :)

PR #116, opened 2022-10-17
feat: Add routes for import/export taxonomy

### LINKED ISSUE #71: Implement taxonomy import from github
- [x] modify parser library to add a labels parameter, that add labels on every nodes (we will use it for branch_name and taxonomy name)
- [x] add an UI to create a project by specifying a new branch name, a taxonomy to import and a description
- [x] add an API point that create a project importing a taxonomy from github (part of it is in #83)

### LINKED ISSUE #83: Add an API to add a new project
Add an API point to create a new project.

- A project has a name (aka branch name) and a taxonomy to work on, it has to be unique.
- it also have a free description field
- taxonomy value should match one of available taxonomies in Open Food Facts.
- the status is "new"
- It creates a node with label PROJECT and with property description, branch_name, taxonomy and status

### LINKED ISSUE #84: Add an API to list projects
after:
- #83

Add an API point that list existing projects.

aadarsh-ram:
### What
- Add routes for importing and dumping a taxonomy

### Related issue(s)
- Fixes #71 
- Fixes #83 
- Fixes #84

### DIFF
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -2,7 +2,6 @@
 Database helper functions for API
 """
 import re
-import os
 import tempfile
 
 import urllib.request                                                           # Sending requests
@@ -100,6 +99,8 @@ class TaxonomyGraph:
         # Close current transaction to use the session variable in unparser
         get_current_transaction().commit()
 
+        self.close_project() # Close the project
+
         # Create unparser object and pass current session to it
         unparser_object = unparser.WriteTaxonomy(get_current_session())
         filename = self.taxonomy_name + '.txt'
@@ -110,7 +111,6 @@ class TaxonomyGraph:
         except:
             raise TaxonomyUnparsingError()
 
-    
     def does_project_exist(self):
         """
         Helper function to check the existence of a project
@@ -150,15 +150,43 @@ class TaxonomyGraph:
             SET n.branch_name = $branch_name
             SET n.description = $description
             SET n.status = $status
+            SET n.created_at = datetime() 
         """
         params = {
             'project_name' : self.project_name,
             'taxonomy_name' : self.taxonomy_name,
             'branch_name' : self.branch_name,
             'description' : description,
-            'status' : 'OPEN'
+            'status' : "OPEN"
         }
         result = get_current_session().run(query, params)
+    
+    def close_project(self):
+        """
+        Helper function to close a Taxonomy Editor project and updates project status as "CLOSED"
+        """
+        query = """
+            MATCH (n:PROJECT)
+            WHERE n.id = $project_name
+            SET n.status = $status
+        """
+        params = {
+            'project_name' : self.project_name,
+            'status' : "CLOSED"
+        }
+        result = get_current_session().run(query, params)
+
+    def list_existing_projects(self):
+        """
+        Helper function for listing all existing projects created in Taxonomy Editor
+        """
+        query = """
+            MATCH (n:PROJECT)
+            WHERE n.status = "OPEN" RETURN n 
+            ORDER BY n.created_at
+        """
+        result = get_current_transaction().run(query)
+        return result
 
     def add_node_to_end(self, label, entry):
         """
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -96,6 +96,23 @@ class TaxonomyGraph:
         else:
             return True
     
+    def check_if_branch_is_unique(self):
+        """
+        Helper function to check uniqueness of GitHub branch
+        """
+        query = """MATCH (n:PROJECT) WHERE n.branch_name = $branch_name RETURN n"""
+        result = get_current_transaction().run(query, {"branch_name" : self.branch_name})
+        if (result.data() == []):
+            return False
+        else:
+            return True
+    
+    def check_branch_name(self):
+        """
+        Helper function to check if a branch name is valid
+        """
+        return normalizer.normalizing(self.branch_name) == self.branch_name
+    
     def create_project(self, description):
         """
         Helper function to create a node with label "PROJECT"
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -4,7 +4,7 @@ Database helper functions for API
 import re
 import urllib.request
 
-from .graph_db import get_current_transaction               # Neo4J transactions helper
+from .graph_db import get_current_transaction, get_current_session               # Neo4J transactions helper
 from openfoodfacts_taxonomy_parser import parser            # Parser for taxonomies
 from openfoodfacts_taxonomy_parser import normalizer        # Normalizing tags
 
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -2,11 +2,14 @@
 Database helper functions for API
 """
 import re
-import urllib.request
+import os
 
-from .graph_db import get_current_transaction, get_current_session               # Neo4J transactions helper
-from openfoodfacts_taxonomy_parser import parser            # Parser for taxonomies
-from openfoodfacts_taxonomy_parser import normalizer        # Normalizing tags
+import urllib.request                                                           # Sending requests
+from .exceptions import TaxnonomyImportError, TaxonomyParsingError              # Custom exceptions
+from .graph_db import get_current_transaction, get_current_session              # Neo4J transactions helper
+
+from openfoodfacts_taxonomy_parser import parser                                # Parser for taxonomies
+from openfoodfacts_taxonomy_parser import normalizer                            # Normalizing tags
 
 class TaxonomyGraph:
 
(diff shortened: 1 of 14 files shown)


backend/editor/entries.py
```diff
@@ -44,6 +49,24 @@ def create_node(self, label, entry, main_language_code):
...
+        parser_object = parser.Parser()
+        try:
+            parser_object(filename, self.branch_name, self.taxonomy_name)
+            return True
+        except:
+            return False
+
+    def import_from_github(self):
+        base_url = "https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-server/main/taxonomies/"
+        filename = ('_'.join(self.taxonomy_name.lower().split()) + '.txt')
```
alexgarel, line 62: small tip, you can use tempfile module to create a tempfile (even with a context manager !)

backend/editor/entries.py
```diff
@@ -44,6 +54,104 @@ def create_node(self, label, entry, main_language_code):
...
+        filename = self.taxonomy_name + '.txt'
+        try:
+            # Parse taxonomy with given file name and branch name
+            unparser_object(filename, self.branch_name, self.taxonomy_name)
+            return filename
+        except:
+            return TaxonomyUnparsingError()
+
+    
+    def check_if_project_exists(self):
```
alexgarel, line 108: picky/personal preference: 

as this return True or False, it's better named `project_exists` 
so that you can read  `if (taxonomy.project_exist())`
When you use check_.... I would expect it to raise instead if it's not ok.

same below: `is_branch_unique`

backend/editor/entries.py
```diff
@@ -44,6 +54,104 @@ def create_node(self, label, entry, main_language_code):
...
+        """
+        base_url = "https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-server/main/taxonomies/"
+        filename = self.taxonomy_name + '.txt'
+        base_url += filename
+        try:
+            # Downloads and creates taxonomy file in current working directory
+            urllib.request.urlretrieve(base_url, filename)
+            status = self.parse_taxonomy(filename) # Parse the taxonomy
+            self.create_project(description) # Creates a "project node" in neo4j
+            os.remove(filename) # Removes taxonomy file after parsing
```
alexgarel, line 85: Great, that's really nice to read :-)

However, here is how to use tempfile, (so that you guarantee no race condition):

up on the file
```
import tempfile
```

```suggestion
            filepath = f"{tmpdir}/{filename}"
            with tempfile.TemporaryDirectory(prefix="taxonomy-") as tmpdir:
                # Downloads and creates taxonomy file in current working directory
                urllib.request.urlretrieve(base_url, filepath)
                status = self.parse_taxonomy(filepath) # Parse the taxonomy
                self.create_project(description) # Creates a "project node" in neo4j
```

no need to clean, the directory is removed at the end.

parser/tests/unit/test_parser_unit.py
```diff
@@ -1,30 +1,47 @@
...
 TEST_TAXONOMY_TXT = str(pathlib.Path(__file__).parent.parent / "data" / "test.txt")
 
 
 def test_normalized_filename():
-    x = parser.Parser()
+
+    # Initialize neo4j
+    uri = os.environ.get("NEO4J_URI", "bolt://localhost:7687")
+    driver = GraphDatabase.driver(uri)
+    session = driver.session()
```
alexgarel, line 17: in case you care, it could have been written as a [pytest fixture](https://docs.pytest.org/en/latest/how-to/fixtures.html) :-)

In `conftest.py`:

```python
import pytest

@pytest.fixture
def graph_session():
    driver = GraphDatabase.driver(uri)
    session = driver.session()
    try:
        yield session
    finally:
        session.close()
```

And then in each test, just get a graph_session parameter, that will give you the session:

```python
def test_normalized_filename(graph_session):
    x = parser.Parser(graph_session)
    ...
    # no need to close
```

alexgarel (approved):
Really a great work.

I propose some improvements, but I approve it so you can merge as soon as you want.

backend/editor/entries.py
aadarsh-ram: Yep @alexgarel, that's actually a good choice. I'll do the same from now on :)

backend/editor/entries.py
aadarsh-ram: Ah, I was creating a tempfile instead of a directory and was facing errors. This is actually great!

parser/tests/unit/test_parser_unit.py
aadarsh-ram: That's cool, I'll refactor it accordingly

aadarsh-ram:
@alexgarel I've added export functionality (Download and Github) endpoints in the last 2 commits. Please review the same.

backend/editor/entries.py
```diff
@@ -78,34 +82,63 @@ def import_from_github(self, description):
...
         unparser_object = unparser.WriteTaxonomy(get_current_session())
         filename = self.taxonomy_name + '.txt'
         try:
             # Parse taxonomy with given file name and branch name
             unparser_object(filename, self.branch_name, self.taxonomy_name)
-            return filename
+            if type_of_export == "download":
+                return filename
+            elif type_of_export == "github":
+                return self.export_to_github(filename)
```
alexgarel, line 117: (Here I'm really picky, but I know you like to learn :-) )

When you add a parameter and the return type is different based on this, this is a good example of a "code smell", that means it is not the best way to factor the code.

To make it better, I would have two different function but calling a common dump_taxonomy function (containing the above code, but the close project part)
```python
    def dump_taxonomy():
         # Create unparser object and pass current session to it
        unparser_object = unparser.WriteTaxonomy(get_current_session())
        filename = self.taxonomy_name + '.txt'   
        try:
            # Parse taxonomy with given file name and branch name
            unparser_object(filename, self.branch_name, self.taxonomy_name)
            return filename
        except Exception:
             raise TaxonomyUnparsingError()

    def file_export(self):
        # Close current transaction to use the session variable in unparser
        get_current_transaction().commit()
        filepath = self.dump_taxonomy()
        return filepath

    def github_export(self):
        # Close current transaction to use the session variable in unparser
        get_current_transaction().commit()
        filepath = self.dump_taxonomy()
        result = self.export_to_github(filepath)
        self.close_project()
        return result
```

You might think it's a bit of repetition, but I think this kind of high level repetition do not harm, as really the logic tend to be a bit different between both cases (and might differ more)

backend/editor/entries.py
```diff
@@ -78,34 +82,63 @@ def import_from_github(self, description):
...
+            raise TaxonomyUnparsingError()
     
-    def check_if_project_exists(self):
+    def export_to_github(self, filename):
+        """
+        Helper function to export a taxonomy to GitHub
+        """
+        query = """MATCH (n:PROJECT) WHERE n.id = $project_name RETURN n.description"""
+        result = get_current_session().run(query, {"project_name": self.project_name})
+        description = result.data()[0]['n.description']
```
alexgarel, line 127: Note: in the future, we could add a link to the taxonomy editor project so that reviewer can check it there :-)

backend/editor/entries.py
```diff
@@ -143,14 +180,44 @@ def create_project(self, description):
...
+        query = """
+            MATCH (n:PROJECT)
+            WHERE n.id = $project_name
+            SET n.status = $status
+        """
+        params = {
+            'project_name' : self.project_name,
+            'status' : "CLOSED"
+        }
+        result = get_current_session().run(query, params)
```
alexgarel, line 208: Picky again: I feel like this is a bit risky to call current session here while all over the place we use current transaction.

One way to do it better might be to have a NewTransactionCtx in graph_db.py defined like:

```python
@contextlib.contextmanager
def NewTransactionCtx():
    """
    Transaction context will set global transaction "txn" for the code in context
    Use it to open a new transaction in current session
    """
    global txn
    with get_current_session().begin_transaction() as _txn:
        txn.set(_txn)
        yield txn
    txn.set(None)
```
and in code upper to close_project, (I prefer there, as it is the code, closing the primary transaction in the first place), use it:
```python3
        with NewTransactionCtx():
            self.close_project()
```

BTW the TransactionCtx could in turn call this, instead of duplicating code :-)

alexgarel (approved):
Really great work @aadarsh-ram 

I made some comment, but you are free to ignore them if you want :-) as it's more perfectionist comments.

backend/editor/entries.py
```diff
@@ -78,34 +82,63 @@ def import_from_github(self, description):
...
-    def export_taxonomy(self):
+    def export_taxonomy(self, type_of_export):
         """
-        Helper function to export a taxonomy for download
+        Helper function to export a taxonomy through download or Github
         """
         # Close current transaction to use the session variable in unparser
         get_current_transaction().commit()
 
+        self.close_project() # Close the project
```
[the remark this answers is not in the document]
aadarsh-ram: Yep, you're right. I'll make this change.

backend/editor/entries.py
```diff
@@ -78,34 +82,63 @@ def import_from_github(self, description):
...
+
         # Create unparser object and pass current session to it
         unparser_object = unparser.WriteTaxonomy(get_current_session())
         filename = self.taxonomy_name + '.txt'
```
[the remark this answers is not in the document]
aadarsh-ram: There's a chance of the files getting mixed up if we change it in the parameters of FileResponse. Instead, I changed the filename that we give to the unparser.

backend/editor/entries.py
aadarsh-ram: That's really informative. Thanks for the tip!

backend/editor/entries.py
aadarsh-ram: Yep, I used the TransactionCtx again and changed everything to use transactions. :)

alexgarel:
For me it's all good @aadarsh-ram

backend/editor/settings.py
```diff
@@ -1,4 +1,8 @@
-# Settings for Neo4J
-
+"""
+Settings for Neo4J
+"""
 import os
+
 uri = os.environ.get("NEO4J_URI", "bolt://localhost:7687")
+access_token = os.environ.get("GITHUB_PAT", "<Add personal access token here>")
```
aadarsh-ram, line 7: @alexgarel Please do let me know how this will be added in production/staging.

aadarsh-ram:
@alexgarel I'll be merging this to the "multilabelnew" branch.

PR #125, opened 2022-11-13
docs: Updated README's and other documentation

aadarsh-ram:
### What
- Updated README.md
- Updated setup-dev.md


teolemon (approved):
That looks great 👍

README.md
```diff
@@ -1,19 +1,31 @@
 # Taxonomy Editor
-Taxonomies are at the 🧡 of Open Food Facts data structure. This project provides an user-friendly editor for editing taxonomies easily.
 
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="https://static.openfoodfacts.org/images/logos/off-logo-horizontal-dark.svg">
+  <source media="(prefers-color-scheme: light)" srcset="https://static.openfoodfacts.org/images/logos/off-logo-horizontal-light.svg">
+  <img height="90" src="https://static.openfoodfacts.org/images/logos/off-logo-horizontal-light.svg">
```
alexgarel, line 6: Also code seems correct, this does not work for me ! 

image

I would say it's because those are external images and github seems to transform them to camo.github … url, screwing them in the process.

So maybe the simplest is to include those svg images in the repository (in the doc/ folder)

aadarsh-ram: Yep @alexgarel, added it. Checked it with light mode too. 

Screenshot

PR #128, opened 2022-11-15
fix: Creation and relationship building using normalized node ID

aadarsh-ram:
### What
- Removing ":" from normalization during creation of a node
- Changed to normalized node IDs during building of a relationship


backend/editor/api.py
```diff
@@ -301,12 +301,12 @@ async def createNode(request: Request, branch: str, taxonomy_name: str):
...
-    taxonomy.create_node(taxonomy.get_label(id), id, main_language)
+    normalized_id = taxonomy.create_node(taxonomy.get_label(id), id, main_language)
     if (taxonomy.get_label(id) == "ENTRY"):
-        taxonomy.add_node_to_end(taxonomy.get_label(id), id)
+        taxonomy.add_node_to_end(taxonomy.get_label(normalized_id), normalized_id)
     else:
-        taxonomy.add_node_to_beginning(taxonomy.get_label(id), id)
-
+        taxonomy.add_node_to_beginning(taxonomy.get_label(normalized_id), normalized_id)
+
```
alexgarel, line 309: Right !

backend/editor/entries.py
```diff
@@ -40,28 +40,31 @@ def create_node(self, label, entry, main_language_code):
...
+            # Normalizing new canonical tag
+            language_code, canonical_tag = entry.split(":", 1)
+            normalised_canonical_tag = normalizer.normalizing(canonical_tag, main_language_code)
+            
+            # Reconstructing and updation of node ID
+            params["id"] = language_code + ':' + normalised_canonical_tag
             params["main_language_code"] = main_language_code
+
+            query.append(f""" SET n.main_language = $main_language_code """) # Required for only an entry
         else:
```
alexgarel, line 57: FYI:

If we want to support stopwords and synonyms there might be more work to be done:
from docs/references/taxonomy-parser.md:
```
2. store each block of *stopwords line* in a node with id `stopwords:IDX`
   and label STOPWORDS,
   where `IDX` is the number of preceding *stopwords lines* in the file
3. store each block of *global synonyms line*
   in a node with id `synonyms:IDX`
   and label SYNONYMS,
   where `IDX` is the number of preceding *global synonyms lines* 
   in the file
```
and it's not the responsability of the frontend to compute this IDX part.

alexgarel (approved):
All good

PR #132, opened 2022-11-22
feat: Update parser to add a neo4j parsing errors node

### LINKED ISSUE #57: Collect parsing errors in a neo4j node
We add logs for line with problem while parsing the taxonomy file (see #37). It's a good first step.

But in the future, the script will be run by the web interface and we may want to show user parsing errors.

We should collect the errors while parsing (and building links) and add them to a special "errors" (or "parsing errors") node.

This node could be used then to display errors.

aadarsh-ram:
### What
- Add all warnings and errors to a node in Neo4j

### Fixes bug(s)
- Fixes #57

### DIFF
--- a/parser/openfoodfacts_taxonomy_parser/parser.py
+++ b/parser/openfoodfacts_taxonomy_parser/parser.py
@@ -20,6 +20,10 @@ class Parser:
 
     def __init__(self, session):
         self.session = session
+        logging.basicConfig(
+            handlers=[logging.FileHandler(filename="parser.log", encoding="utf-8", mode="w")],
+            level=logging.INFO,
+        )
 
     def create_headernode(self, header, multi_label):
         """Create the node for the header"""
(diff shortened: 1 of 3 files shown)


parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -14,16 +14,36 @@ def ellipsis(text, max=20):
     """Cut a text adding eventual ellipsis if we do not display it fully"""
     return text[:max] + ("..." if len(text) > max else "")
 
+class ParserConsoleLogger:
```
alexgarel, line 18: Really good idea to put it in a separate class :+1:

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -15,11 +15,37 @@ def ellipsis(text, max=20):
...
+    def __init__(self):
+        self.parsing_warnings = []  # Stores all warning logs
+        self.parsing_errors = []  # Stores all error logs
+
+        logging.basicConfig(
+            handlers=[logging.StreamHandler()],
+            level=logging.INFO,
+        )
```
alexgarel, line 26: We must keep this out of here.
It's up to the caller to configure this, you really don't want this here.

You should move it back to the `if __name__ == "__main__"` part.

alexgarel (approved):
That's perfect, appart from the logging config that should be restored as it was.

parser/openfoodfacts_taxonomy_parser/parser.py
aadarsh-ram: Since this class is used for all the logging operations, I wanted to do the configuration when the ParserLogger class was created.

If we move it to the __main__ part, the config will only be done when the parser is run as a separate file. It will not be configured, when the `entries.py` file creates the Parser class.

alexgarel: Yes but that's not the role of this class to setup logs for entries.py !

The FastAPI server should setup logs for itself and no sub part should change that.

That is, it's not the role of a sub-part of the code to setup loging, it's the role of the main program.

For the FastAPI server let setup log in api.py if you want, or in a settings.py file.

alexgarel: I found a citation to explain it better: https://docs.python.org/3.8/howto/logging.html#configuring-logging

> It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3.8/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

aadarsh-ram: Yep, understood @alexgarel. Thanks for the help!

aadarsh-ram:
@alexgarel I'll be merging this PR now.

PR #133, opened 2022-11-28
build: fix confs

alexgarel:
Fix configuration errors.


aadarsh-ram (approved):
LGTM. Works in local with no issues👍

PR #137, opened 2022-12-02
fix: Normalize '-' characters + fix: Convert to snake_case

aadarsh-ram:
### What
- Branch name containing hyphens were not normalized
- All API functions and variables converted to snake case for readability


alexgarel (approved):
:clap:

PR #141, opened 2022-12-02
fix: nginx config allows app page reload

alexgarel:
we have to redirect unknown url to index.html so that react app can handle them.


aadarsh-ram (approved):
LGTM

PR #152, opened 2022-12-22
feat: Added functions for reopening projects and fetching closed projects

aadarsh-ram:
### What
- Added helper function for reopening a project
- Added helper function for fetching closed projects
- Added two new routes in API for the same

### DIFF
--- a/backend/editor/api.py
+++ b/backend/editor/api.py
@@ -114,14 +114,25 @@ async def pong(response: Response):
     return {"ping": "pong @ %s" % pong}
 
 
-@app.get("/projects")
-async def list_all_projects(response: Response):
+@app.get("/open-projects")
+async def list_all_open_projects(response: Response):
     """
     List all open projects created in the Taxonomy Editor
     """
-    # Listing all projects doesn't require a taoxnomy name or branch name
+    # Listing all projects doesn't require a taxonomy name or branch name
     taxonony = TaxonomyGraph("", "")
-    result = list(taxonony.list_existing_projects())
+    result = list(taxonony.list_all_open_projects())
+    return result
+
+
+@app.get("/closed-projects")
+async def list_all_closed_projects(response: Response):
+    """
+    List all closed projects created in the Taxonomy Editor
+    """
+    # Listing all projects doesn't require a taxonomy name or branch name
+    taxonony = TaxonomyGraph("", "")
+    result = list(taxonony.list_all_closed_projects())
     return result
 
 
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -240,14 +240,38 @@ class TaxonomyGraph:
         """
         params = {"project_name": self.project_name, "status": "CLOSED"}
         get_current_transaction().run(query, params)
+    
+    def reopen_project(self):
+        """
+        Helper function to reopen a closed Taxonomy Editor project and updates project status as "OPEN"
+        """
+        query = """
+            MATCH (n:PROJECT)
+            WHERE n.id = $project_name
+            SET n.status = $status
+        """
+        params = {"project_name": self.project_name, "status": "OPEN"}
+        get_current_transaction().run(query, params)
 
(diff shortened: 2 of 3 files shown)


backend/editor/api.py
```diff
@@ -114,14 +114,25 @@ async def pong(response: Response):
...
+
+@app.get("/closed-projects")
+async def list_all_closed_projects(response: Response):
+    """
+    List all closed projects created in the Taxonomy Editor
+    """
+    # Listing all projects doesn't require a taxonomy name or branch name
+    taxonony = TaxonomyGraph("", "")
+    result = list(taxonony.list_all_closed_projects())
     return result
```
alexgarel, line 136: I'm ok with this, but we could have also simply added a "status" filter, it would enable more operations in the future.

It seems to me that we should at least factor the code in entries.py to have a single method, with a status argument (which can be None, to get everything), because it avoids repeating too much same code (DRY).

backend/editor/entries.py
```diff
@@ -253,6 +266,18 @@ def list_existing_projects(self):
         result = get_current_transaction().run(query)
         return result
 
+    def list_all_closed_projects(self):
```
alexgarel, line 269: As said above, I would prefer that we factor  list_all_open_projects and list_all_closed_projects into a single method. What do you think ?

backend/editor/entries.py
```diff
@@ -241,7 +241,20 @@ def close_project(self):
         params = {"project_name": self.project_name, "status": "CLOSED"}
         get_current_transaction().run(query, params)
 
-    def list_existing_projects(self):
+    def reopen_project(self):
```
alexgarel, line 244: there again, do we want a reopen_project method, or just a set_status method ?

If there are a lot of pre/post processing specific to that case, it make sense to have a specific method (for example, if we were sending an email, etc.) but if not ain't it better to have a generic "set_status" method ?

alexgarel (approved):
I approve because my comments are opinionated. So feel free to merge or to refactor first ;-)

backend/editor/entries.py
aadarsh-ram: Yep, I think it makes sense to club it into a single method. Using query params, it'll be easier to list projects with a single endpoint as well.

backend/editor/entries.py
aadarsh-ram: Yep, it should be better. So, I changed it to a set_status() method and used it for closing a project as well.

aadarsh-ram:
@alexgarel There seems to be a error with the checks workflow: https://github.com/openfoodfacts/taxonomy-editor/actions/runs/3789170499/jobs/6442677225. Please do check it out.

backend/editor/entries.py
```diff
@@ -229,28 +229,37 @@ def create_project(self, description):
...
+        query = ["""MATCH (n:PROJECT)\n"""]
+        if status == "OPEN":
+            # List only the open projects
+            query.append("""WHERE n.status = "OPEN"\n""")
+        elif status == "CLOSED":
+            # List only the closed projects
+            query.append("""WHERE n.status = "CLOSED"\n""")
+        elif status == "ALL":
+            # List all the projects, irrespective of their status
+            pass
```
alexgarel, line 257:
```suggestion
        params = {}
        if status != "ALL":
            # List only projects matching status
            query.append("""WHERE n.status = $status\n""")
            params["status"] = status
```

backend/editor/entries.py
```diff
@@ -229,28 +229,37 @@ def create_project(self, description):
...
+
+        query.append("""RETURN n\n""")
+        query.append("""ORDER BY n.created_at\n""")  # Sort by creation date
+
+        result = get_current_transaction().run("".join(query))
```
alexgarel, line 262:
```suggestion
        result = get_current_transaction().run("".join(query), params)
```

alexgarel (approved):
There again my comments are only enhancements comments, so I already approve.

PR #153, opened 2022-12-22
build: Remove redundant parser checks

aadarsh-ram:
### What
- The Makefile for Taxonomy Editor already contains checks for parser
- Removed separate parser checks from pull_request.yml


alexgarel (approved):
LGTM

alexgarel:
Thanks @aadarsh-ram, this will save some CO2 emissions :-)

Issue #154, opened 2022-12-23
Only show some languages in interface


alexgarel:
### What

Currently we display every language on every entries in the taxonomy editor.

This is really cumbersome as one might only want to eg. translate in Polish from english and maybe spanish.

### Proposed solution

On an entry, only translations in a selected list of languages are displayed. 

A dropdown makes it possible to shown languages. 
This parameter is persisted in local storage, so that if I come back, it's still there.

At any moment I can display a language (maybe because I wont to see how it's handle there) then hide it again… always through this same "displayed language list".

If I add a language ([+] button next to "Translations"), and add an entry for currently not displayed language, it is automatically selected for display.

Note: The main entry language is always displayed, to avoid perception that a page is empty.

WaqarTabish2807:
Hey Alex :))

There is always an issue as "Failed to fetch" in the editor. Can you please help me with it?

PR #159, opened 2022-12-31
feat: show only some translations in the interface

diivi:
### What
- add visibility toggle to languages
- hide all languages by default
- store user interacted languages in localstorage.

### Screenshot
visibility_toggle

### Fixes bug(s)
#154

### DIFF
--- a/taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
+++ b/taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
@@ -7,6 +7,8 @@ import DialogTitle from '@mui/material/DialogTitle';
 import { useEffect, useState } from "react";
 import AddBoxIcon from '@mui/icons-material/AddBox';
 import DeleteOutlineIcon from '@mui/icons-material/DeleteOutline';
+import VisibilityIcon from '@mui/icons-material/Visibility';
+import VisibilityOffIcon from '@mui/icons-material/VisibilityOff';
 import ISO6391 from 'iso-639-1';
 
 /**
@@ -19,6 +21,7 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
     const [openDialog, setOpen] = useState(false); // Used for Dialog component
     const [newLanguageCode, setNewLanguageCode] = useState(''); // Used for storing new LC from Dialog
     const [isValidLanguageCode, setisValidLanguageCode] = useState(false); // Used for validating a new LC
+    const [shownTranslationLanguages, setShownTranslationLanguages] = useState([]); // Used for storing LC's that are to be shown
 
     // Helper functions for Dialog component
     const handleClose = () => { setOpen(false); }
@@ -27,6 +30,7 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
     // Used for addition of a translation language
     const handleAddTranslation = (key) => {
         const newRenderedTranslations = [...renderedTranslations, {'languageCode' : key, 'tags' : []}]
+        handleToggleTranslationVisibility(key)
         setRenderedTranslations(newRenderedTranslations);
         key = 'tags_' + key; // LC must have a prefix "tags_"
         const uuidKey = key + '_uuid' // Format for the uuid
@@ -58,9 +62,18 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
         setOpen(false);
     }
 
+    const handleToggleTranslationVisibility = (key) => {
+        const newShownTranslationLanguages = shownTranslationLanguages.includes(key) ? shownTranslationLanguages.filter(obj => !(key === obj)) : [...shownTranslationLanguages, key]
+        setShownTranslationLanguages(newShownTranslationLanguages);
+        localStorage.setItem('shownTranslationLanguages', JSON.stringify(newShownTranslationLanguages));
+    }
+
     // Changes the translations to be rendered
     // Dependent on changes occuring in "nodeObject"
     useEffect(() => {
+        // get shown translation languages from local storage
+        const shownTranslationLanguages = JSON.parse(localStorage.getItem('shownTranslationLanguages')) || [];
+        setShownTranslationLanguages(shownTranslationLanguages);
 
         // Main langauge tags are considered separately, since they have to be rendered first
         const mainLangTags = []
@@ -287,6 +300,9 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
                                 <IconButton onClick={() => handleDeleteTranslation(lang)}>
                                     <DeleteOutlineIcon />
                                 </IconButton>
+                                <IconButton onClick={() => handleToggleTranslationVisibility(lang)}>
+                                    {shownTranslationLanguages.includes(lang) ? <VisibilityIcon /> : <VisibilityOffIcon />}
+                                </IconButton>
                             </Stack>
                             {/* Render all related tags */}
                             {
@@ -294,6 +310,7 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
                                     const index = tagObj['index']
                                     const tag = tagObj['tag']
                                     return (
+                                        shownTranslationLanguages.includes(lang) &&
                                         <Stack key={index} sx={{ml: 2}} direction="row" alignItems="center">
                                             <TextField 
                                                 size="small" 
--- a/taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
+++ b/taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
@@ -21,7 +21,7 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
     const [newLanguageCode, setNewLanguageCode] = useState(''); // Used for storing new LC from Dialog
     const [languageAction, setLanguageAction] = useState('invalid'); // Used for storing state of action to be performed on LC (add/show/invalid)
     const [shownTranslationLanguages, setShownTranslationLanguages] = useState([]); // Used for storing LC's that are to be shown
-    const [languageOptions, setLanguageOptions] = useState([]); // Used for storing all possible LC's
+    const [addedLanguageCodes, setAddedLanguageCodes] = useState([]); // Used for storing LC's that are added to the nodeObject
 
     // Helper functions for Dialog component
     const handleClose = () => { setOpen(false); }
@@ -119,18 +119,8 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
             }
         })
 
-        const languageOptions = []
-        // All language options for add dropdown (already added languages have a 👁 icon)
-        ISO6391.getAllCodes().forEach((languageCode) => {
-            if (addedLanguageCodes.includes(languageCode)) {
-                languageOptions.push(ISO6391.getName(languageCode) + "👁️")
-            }
-            else {
-                languageOptions.push(ISO6391.getName(languageCode))
-            }
-        })
         // Set states
-        setLanguageOptions(languageOptions)
+        setAddedLanguageCodes(addedLanguageCodes);
         setMainLangRenderedTranslations(mainLangTags);
         setRenderedTranslations(otherLangTags);
     }, [nodeObject, shownTranslationLanguages]);
@@ -359,7 +349,7 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
                 </DialogContentText>
                 <Autocomplete
                     sx={{mt: 2}}
-                    options={languageOptions}
+                    options={ISO6391.getAllNames()}
                     onChange={(e,language) => {
                         if (!language) language = '';
                         if (language.includes('👁️')) language = language.slice(0,-3) // exclude the eye emoji for validation
@@ -371,6 +361,16 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
                         if (isValidLanguage  && !isDuplicateLanguage) {setLanguageAction("add")}
                         else if (isValidLanguage && !isAlreadyShown && isDuplicateLanguage) {setLanguageAction("show")}
                         else {setLanguageAction("invalid")}
+                    }}
+                    renderOption={(props, option) => {
+                        const languageCode = ISO6391.getCode(option)
+                        const isAlreadyShown = addedLanguageCodes.includes(languageCode)
+                        // set color of text to green if the language is already shown
+                        return (
+                            <span {...props} style={{ color: isAlreadyShown ? 'green' : 'black' }}>
+                                {option}
+                            </span>
+                        )
                     }}
                     renderInput={(params) =>
                         <TextField
--- a/taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
+++ b/taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
@@ -7,21 +7,20 @@ import DialogTitle from '@mui/material/DialogTitle';
 import { useEffect, useState } from "react";
 import AddBoxIcon from '@mui/icons-material/AddBox';
 import DeleteOutlineIcon from '@mui/icons-material/DeleteOutline';
-import VisibilityIcon from '@mui/icons-material/Visibility';
 import VisibilityOffIcon from '@mui/icons-material/VisibilityOff';
 import ISO6391 from 'iso-639-1';
 
 /**
  * Sub-component for rendering translation of an "entry"
 */
-const ListTranslations = ({ nodeObject, setNodeObject }) => {
+const ListTranslations = ({ nodeObject, setNodeObject, shownTranslationLanguages, setShownTranslationLanguages }) => {
 
     const [renderedTranslations, setRenderedTranslations] = useState([]) // Stores state of all tags
     const [mainLangRenderedTranslations, setMainLangRenderedTranslations] = useState([]) // Stores state of main language's tags
     const [openDialog, setOpen] = useState(false); // Used for Dialog component
     const [newLanguageCode, setNewLanguageCode] = useState(''); // Used for storing new LC from Dialog
-    const [isValidLanguageCode, setisValidLanguageCode] = useState(false); // Used for validating a new LC
-    const [shownTranslationLanguages, setShownTranslationLanguages] = useState([]); // Used for storing LC's that are to be shown
+    const [languageAction, setLanguageAction] = useState('invalid'); // Used for storing state of action to be performed on LC (add/show/invalid)
+    const [languageOptions, setLanguageOptions] = useState([]); // Used for storing all possible LC's
 
     // Helper functions for Dialog component
     const handleClose = () => { setOpen(false); }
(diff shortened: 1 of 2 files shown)


alexgarel:
@diivi that's really cool.

Only bug for me:
* language preferences are not saved when I navigate to another node. It seems that languages are saved only when I hit the "submit button". For me it should be saved every time you make a change to language selection (hide one or add one). It should not be correlated to "submit button"

@nobeeakon would you have time to get an eye on this PR ?

aadarsh-ram (changes requested):
Hey @diivi, I added some comments. Thanks for the contribution!

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -314,38 +340,44 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
                     )
                 } )
             }
-            {/* Dialog box for adding translations */}
+            {/* Button to add language to the list of shown languages or show the language if it already exists */}
+            <Button sx={{mt: 2}} onClick={handleOpen}>Add/Show More Languages ({renderedTranslations.length - shownTranslationLanguages.length} Hidden)</Button>
```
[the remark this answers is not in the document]
diivi: My bad, should be fixed now.

alexgarel:
I see you use a color to show existing languages ! That's cool you managed to do it, but I would keep a small icon (or use bold) to help color blind users.

alexgarel:
Also a small possible improvement (that I understand using your UI):

Imagine I selected Czech and Quechua as languages. When I go to a page were there is no Quechua translation, I have to click "add language" while this would just add the Quechua language without any entry.

Why don't we display the Quechua language without any entry (so that I can add my entries).

image

This did not make sense before, when every existing entry were visible, but now it makes more sense to display empty entries for selected languages.

alexgarel:
Kudos @diivi that's really cool :-)

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -328,24 +374,37 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
                     }}
-                    renderInput={(params) => 
+                    renderOption={(props, option) => {
+                        const languageCode = ISO6391.getCode(option)
+                        const isAlreadyShown = addedLanguageCodes.includes(languageCode)
+                        // set color of text to green if the language is already shown
+                        return (
+                            <span {...props} style={{ color: isAlreadyShown ? 'green' : 'black' }}>
+                                {isAlreadyShown ? <b>{option}</b> : option}
+                            </span>
```
alexgarel, line 391: If I'm correct, it's not if it's already shown but if it already have a translation.

So you may change comment and variable name.

diivi: True, will do.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -328,24 +374,37 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
-                    disabled={!isValidLanguageCode}
-                    onClick={() => {handleAddTranslation(newLanguageCode)}}>
-                        Add
-                </Button>
+                { languageAction !== "invalid" && 
+                    <Button
+                        onClick={() => {languageAction === "add" ? handleAddTranslation(newLanguageCode) : handleToggleTranslationVisibility(newLanguageCode)}}>
+                            {languageAction === "add" ? "Add" : "Show" }
+                    </Button>
+                }
```
alexgarel, line 407: If languageAction is invalid, I think we could add a warning message instead of the button, so that user understand better the dynamic.

```suggestion
                { languageAction !== "invalid" && 
                    <Button
                        onClick={() => {languageAction === "add" ? handleAddTranslation(newLanguageCode) : handleToggleTranslationVisibility(newLanguageCode)}}>
                            {languageAction === "add" ? "Add" : "Show" }
                    </Button> 
                    ||
                    <Button disabled>Already selected</Button>
                }
```

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
diivi: Actually, it will be green for one more case,
when the user added a language, but did not add a translation yet, so I am renaming it to `presentInObjectNode`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
diivi: image
The invalid state itself can have multiple reasons 😅 , like not selecting any language.
Should I just put `Select hidden language` here

alexgarel: it's ok, like this :-)

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -58,13 +65,23 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
         setOpen(false);
     }
 
+    // Used for toggling visibility of a translation language
+    const handleToggleTranslationVisibility = (key) => {
+        const newShownTranslationLanguages = shownTranslationLanguages.includes(key) ? shownTranslationLanguages.filter(obj => !(key === obj)) : [...shownTranslationLanguages, key]
+        setShownTranslationLanguages(newShownTranslationLanguages);
+        newShownTranslationLanguages.sort()
+        localStorage.setItem('shownTranslationLanguages', JSON.stringify(newShownTranslationLanguages))
+        setOpen(false);
```
nobeeakon, line 74: nit: could we rename this `setOpen` to something more explicit (maybe `isOpenDialog`, `setIsOpenDialog` or something)

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -58,13 +65,23 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
         setOpen(false);
     }
 
+    // Used for toggling visibility of a translation language
+    const handleToggleTranslationVisibility = (key) => {
+        const newShownTranslationLanguages = shownTranslationLanguages.includes(key) ? shownTranslationLanguages.filter(obj => !(key === obj)) : [...shownTranslationLanguages, key]
+        setShownTranslationLanguages(newShownTranslationLanguages);
+        newShownTranslationLanguages.sort()
+        localStorage.setItem('shownTranslationLanguages', JSON.stringify(newShownTranslationLanguages))
```
nobeeakon, line 73: nit: maybe have this key name (`shownTranslationLanguages`) as a const outside of the component, just so is easier to keep track of it

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -102,10 +120,26 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
         // Set states
+        setAddedLanguageCodes(addedLanguageCodes);
         setMainLangRenderedTranslations(mainLangTags);
         setRenderedTranslations(otherLangTags);
-    }, [nodeObject]);
+    }, [nodeObject, shownTranslationLanguages]);
+
+    useEffect(() => {
+        // get shown translation languages from local storage
+        const shownTranslationLanguages = JSON.parse(localStorage.getItem('shownTranslationLanguages')) || [];
```
nobeeakon, line 140: please be aware that `JSON.parse` can throw an error. (you can test it in the console using: `JSON.parse('}')`). Also, as the local storage is exposed people can mess with it

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -58,13 +65,23 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
         setOpen(false);
     }
 
+    // Used for toggling visibility of a translation language
+    const handleToggleTranslationVisibility = (key) => {
+        const newShownTranslationLanguages = shownTranslationLanguages.includes(key) ? shownTranslationLanguages.filter(obj => !(key === obj)) : [...shownTranslationLanguages, key]
+        setShownTranslationLanguages(newShownTranslationLanguages);
+        newShownTranslationLanguages.sort()
```
nobeeakon, line 72: do we need the `.sort()`? if so can we do it before setting the state? please be aware that .sort() modifies the array that was passed to the state

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -314,12 +352,20 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
                     )
                 } )
             }
-            {/* Dialog box for adding translations */}
+            {/* Button to add language to the list of shown languages or show the language if it already exists */}
+            <Button sx={{mt: 2}} onClick={handleOpen}>
+                Add/Show More Languages 
+                ({
+                    addedLanguageCodes.length - 
+                    addedLanguageCodes.filter(element => new Set(shownTranslationLanguages).has(element)).length
```
nobeeakon, line 360: nit: this `Set` can be `shownTranslationLanguages.includes(element)` . Also, if possible, let's rename `element` with something a bit more descriptive (`languageCodeItem`  or something)

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -18,7 +19,9 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
     const [mainLangRenderedTranslations, setMainLangRenderedTranslations] = useState([]) // Stores state of main language's tags
     const [openDialog, setOpen] = useState(false); // Used for Dialog component
     const [newLanguageCode, setNewLanguageCode] = useState(''); // Used for storing new LC from Dialog
-    const [isValidLanguageCode, setisValidLanguageCode] = useState(false); // Used for validating a new LC
+    const [languageAction, setLanguageAction] = useState('invalid'); // Used for storing state of action to be performed on LC (add/show/invalid)
```
nobeeakon, line 22: wondering if we could remove the `languageAction`. Or if too complex, at least do not allow the `invalid` state to happen, if that's too complex, guess it's ok as is.  
Let's say, for example, if we replace  the autocomplete with  a <select /> (or it's Material-UI equivalent), then only valid options can be selected by the user. 
Also, not understanding why 'invalid' is the default state, like why is it invalid even before doing anything.
Besides that, wondering if we could just have 'show' instead of 'show' and 'add'. I may be wrong, but my guess is that  from the user perspective there's not much difference, my understanding here is that it would make the user to first 'add' and then 'show' for a particular language? I mean, if the difference between both is not relevant to the user, can we hide it?

diivi: 1. The state is invalid in the beginning to hide the submission button (hide/show) until a valid selection is made.
2. **I think a good reason for keeping the button texts as show and add could be to tell the user that they are adding a node to a taxonomy (everyone will see it) vs when they are just showing a node (only added to their local preferences).**
This is why I did a three state toggle in the first place.
3. Otherwise, the state can toggle between show and invalid (true/false)

nobeeakon: May be missing something , but from a user perspective it seems that we are only adding allowing the user to show more languages, and seems to me that the difference between add and show is just if the language already has information, if that's the case I think we can hide that diference to the user. If that's not the case I'd worry we are mixing different features in the same dialog.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -328,24 +378,37 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
                         if (!language) language = '';
                         setNewLanguageCode(ISO6391.getCode(language));
                         const isValidLanguage = ISO6391.validate(ISO6391.getCode(language))
+                        const isAlreadyShown = shownTranslationLanguages.includes(ISO6391.getCode(language))
                         const isDuplicateLanguage = renderedTranslations.some(el => (el.languageCode === ISO6391.getCode(language))) || 
                                                 nodeObject.main_language === ISO6391.getCode(language)
-                        if (isValidLanguage && !isDuplicateLanguage) {setisValidLanguageCode(true)}
-                        else {setisValidLanguageCode(false)}
+                        if (isValidLanguage  && !isDuplicateLanguage) {setLanguageAction("add")}
```
nobeeakon, line 384: please replace the Autocomplete with [Select](https://mui.com/material-ui/react-select/). Autocomplete are ok when you allow free text (options + something else), here we want the user to only be able to select one of any of the valid languages.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -314,12 +356,20 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
+                } Hidden)
+            </Button>
+            {/* Dialog box for adding languages to the list of shown languages */}
             <Dialog open={openDialog} onClose={handleClose}>
-                <DialogTitle>Add a language</DialogTitle>
+                <DialogTitle>Add or Show another language</DialogTitle>
                 <DialogContent>
                 <DialogContentText>
-                    Enter the two letter language code for the language to be added.
+                    Enter the two letter language code for the language to be shown or added.
```
nobeeakon, line 372: Is this a valid message? It seems to me that we can just remove it

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -55,16 +63,25 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
             delete newNodeObject[uuidKey];
             return newNodeObject
         })
-        setOpen(false);
+        setOpenDialog(false);
+    }
+
+    // Used for toggling visibility of a translation language
+    const handleToggleTranslationVisibility = (key) => {
+        const newShownTranslationLanguages = shownTranslationLanguages.includes(key) ? shownTranslationLanguages.filter(obj => !(key === obj)) : [...shownTranslationLanguages, key]
```
nobeeakon, line 71: nit: `filter(obj => (key !== obj))`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -55,16 +63,25 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
+        setShownTranslationLanguages(newShownTranslationLanguages);
+        localStorage.setItem(translationVisibilityPreferenceKey, JSON.stringify(newShownTranslationLanguages))
+        setOpenDialog(false);
+        setLanguageAction('invalid')
```
nobeeakon, line 75: think  these two (closeDialog, set to invalid)  shouldn't be here. These is used for the dialog (we can have a handler for the `done` action of the dialog that takes care of closing it and calling this function here), but also when hidding something.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -102,10 +120,30 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
         setMainLangRenderedTranslations(mainLangTags);
         setRenderedTranslations(otherLangTags);
-    }, [nodeObject]);
+    }, [nodeObject, shownTranslationLanguages]);
+
+    useEffect(() => {
+        // get shown translation languages from local storage
+        try{
+            const shownTranslationLanguages = JSON.parse(localStorage.getItem(translationVisibilityPreferenceKey)) || []; 
+            setShownTranslationLanguages(shownTranslationLanguages);
```
nobeeakon, line 142: we may want to check if all are valid languages (as people can mess around with the local storage)

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -16,17 +17,24 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
     const [renderedTranslations, setRenderedTranslations] = useState([]) // Stores state of all tags
     const [mainLangRenderedTranslations, setMainLangRenderedTranslations] = useState([]) // Stores state of main language's tags
-    const [openDialog, setOpen] = useState(false); // Used for Dialog component
+    const [openDialog, setOpenDialog] = useState(false); // Used for Dialog component
     const [newLanguageCode, setNewLanguageCode] = useState(''); // Used for storing new LC from Dialog
-    const [isValidLanguageCode, setisValidLanguageCode] = useState(false); // Used for validating a new LC
+    const [languageAction, setLanguageAction] = useState('invalid'); // Used for storing state of action to be performed on LC (add/show/invalid)
+    const [shownTranslationLanguages, setShownTranslationLanguages] = useState([]); // Used for storing LC's that are to be shown
+    const [addedLanguageCodes, setAddedLanguageCodes] = useState([]); // Used for storing LC's that are added to the nodeObject
+    const translationVisibilityPreferenceKey = "shownTranslationLanguages" // Used for storing LC's that are to be shown in localStorage
```
nobeeakon, line 25: this one can live outside of the component (like before `const ListTranslations = ({ nodeObject, setNodeObject }) => {`). Please update comment, to something like: `// local storage key for languages to be shown`

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -16,17 +17,24 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
 
     const [renderedTranslations, setRenderedTranslations] = useState([]) // Stores state of all tags
     const [mainLangRenderedTranslations, setMainLangRenderedTranslations] = useState([]) // Stores state of main language's tags
-    const [openDialog, setOpen] = useState(false); // Used for Dialog component
+    const [openDialog, setOpenDialog] = useState(false); // Used for Dialog component
     const [newLanguageCode, setNewLanguageCode] = useState(''); // Used for storing new LC from Dialog
-    const [isValidLanguageCode, setisValidLanguageCode] = useState(false); // Used for validating a new LC
+    const [languageAction, setLanguageAction] = useState('invalid'); // Used for storing state of action to be performed on LC (add/show/invalid)
+    const [shownTranslationLanguages, setShownTranslationLanguages] = useState([]); // Used for storing LC's that are to be shown
+    const [addedLanguageCodes, setAddedLanguageCodes] = useState([]); // Used for storing LC's that are added to the nodeObject
```
nobeeakon, line 24: can we remove this state, not clearly understanding the difference between this one and `shownTranslations...` Sounds like this could be derived from the nodeObject

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
diivi: Yup, seems redundant.

alexgarel:
This was split in two PRs, notably https://github.com/openfoodfacts/taxonomy-editor/pull/185

Issue #156, opened 2022-12-23
Add route for fetching parsing errors for a project


aadarsh-ram:
During parsing of a taxonomy, all the parsing errors are stored in a separate node with the label "ERRORS" (See function over [here](https://github.com/openfoodfacts/taxonomy-editor/blob/main/parser/openfoodfacts_taxonomy_parser/parser.py#L444)). For the frontend to display the same, these errors need to be accessible through the Taxonomy Editor API.

So, the following needs to be done to enable support to the frontend:
- A function with a neo4j query to fetch parsing errors for a particular project, must be created in [entries.py](https://github.com/openfoodfacts/taxonomy-editor/blob/main/backend/editor/entries.py).
- Add a GET route (maybe `/errors`) in [api.py](https://github.com/openfoodfacts/taxonomy-editor/blob/main/backend/editor/api.py) and it must call the above function and return the results.

diivi:
I have started working on this. Please assign the issue to me

alexgarel:
I suggest naming the route "/parsing_errors" as those are parsing errors.

PR #163, opened 2023-01-03
feat: parsing error endpoint

diivi:
### What
- Add an endpoint to get all parsing errors related to a project
- Modify project list endpoint to include error_count for every project.

### Fixes bug(s)
- #156


backend/editor/entries.py
```diff
@@ -244,10 +244,13 @@ def close_project(self):
...
         """
         query = """
             MATCH (n:PROJECT)
-            WHERE n.status = "OPEN" RETURN n
+            OPTIONAL MATCH (error_node:ERRORS {{branch_name: n.branch_name, id: n.id}})
+            WHERE n.status = "OPEN"
+            RETURN n, size(error_node.errors) AS error_count
             ORDER BY n.created_at
         """
         result = get_current_transaction().run(query)
```
alexgarel, line 256: Here we should transform the result to include our error count in result:

```suggestion
        result = get_current_transaction().run(query)
        # pack errors count in result
        result = [dict(node, errors_count=num_errors) for node, num_errors in result]

```

diivi: Ah, great! I was trying to do this in the query itself 😅

alexgarel (approved):
Great @diivi !

alexgarel:
@diivi merged with main, as there where conflicting changes from @aadarsh-ram 

I also saw my suggestion was not correct (in fact previously list where already nested, so I kept the nesting…)

alexgarel:
First merged request @diivi, kudos :tada: !

PR #172, opened 2023-01-11
feat: CI formatter

nobeeakon:
### What
- In the front end, added prettier and husky. To format code before every commit.  Also runned prettier so this PR has loads of changes, but must of them are just code format changes


alexgarel (changes requested):
@nobeeakon I totally agree with using a linter. That's really cool :-)

That said we have:
- to add an entry to lint and check in Makefile, that use docker, it's to be consistent with the rest of our projects
- add a github action that checks we have linted (not everyone will configure the pre-commit hook), it's simple to setup using the make target
- document how to use it (especially for those not using docker, hence not able to use the make target)

Of course I can help with all of that, but you should explain to me the command to run as the pre-commit is a bit cryptic to me, and I imagine there are some installation step to run ? (like npx prepare ?).
If you can document how to use it, I think I would be able to make the Makefile target and the github action !

alexgarel:
@nobeeakon I added a make target, that makes the checks also run in github actions.

alexgarel (approved):
@nobeeakon please, before merging, add a documentation for devs who wants to use husky (quick setup and a pointers to the package docs)

doc/introduction/setup-dev.md
```diff
@@ -1,5 +1,8 @@
 # Setup a dev environment
 
+- [frontend doc](../../taxonomy-editor-frontend/README.md)
+- [backend doc](../../backend/README.md)
```
alexgarel, line 4: :+1:

PR #182, opened 2023-01-18
fix: use async neo4j

### LINKED ISSUE #166: Adding a children blocks the API server
### What
If we add a children to a node the server, it makes the server unresponsive (infinite loop ?)

Related to:
- #181

Blocks:
- #162

### LINKED ISSUE #181: Neo4J crashing when adding a child
### What
We have a problem experimented in staging where neo4j crashes without a log.

It happens when adding a child to an entry.
You have to remove and recreate the docker container for it to start again (just restarting is not enough)

### Steps to reproduce the behavior:
1. Go to an entry in a big enough taxonomy (like allergens)
2. go to root nodes and choose a node to edit it
3. click on the button to add a child
4. application is not responding any more.
5. behind the scene:
   * neo4j wont respond to any query anymore 
   * even a restart does not make it back (maybe because of some lock, but again neo4j logs are mute)
   * you have to remove / recreate the container

### Expected behavior
This should not crash :grimacing:

alexgarel:
### What

There are 3 commits in  this PR (better review them individually):

* build: handle neo4j memory configration
  + upgrade neo4j and some libs
  + enable reload in dev mode
   + avoid test failing because of neo4j running

* fix: fix bugs in parser
  * tracing errors did raise an exception if there were parameters (added a test)
  * better handling of is_before (avoid class attribute)

* fix: use async neo4j in entries.py  
  Related issues:
  - Fixes #181
  - Fixes #166 

The last one is a big one. I tested around and it seems ok, but I can't be sure I didn't break anything.

We **really need** to add tests to this project !

### DIFF
--- a/backend/editor/api.py
+++ b/backend/editor/api.py
@@ -152,7 +152,7 @@ async def list_all_projects(response: Response, status: Optional[StatusFilter] =
     """
     # Listing all projects doesn't require a taxonomy name or branch name
     taxonomy = TaxonomyGraph("", "")
-    result = taxonomy.list_projects(status)
+    result = await taxonomy.list_projects(status)
     return result
 
 
@@ -400,10 +400,7 @@ async def edit_entry(request: Request, branch: str, taxonomy_name: str, entry: s
     """
     taxonomy = TaxonomyGraph(branch, taxonomy_name)
     incoming_data = await request.json()
-    log.info("start entry update")
     updated_entry = await taxonomy.update_nodes("ENTRY", entry, incoming_data)
-    log.info("end entry update")
-    log.info(f"return {updated_entry}")
     return updated_entry
 
 
@@ -416,10 +413,7 @@ async def edit_entry_children(request: Request, branch: str, taxonomy_name: str,
     """
     taxonomy = TaxonomyGraph(branch, taxonomy_name)
     incoming_data = await request.json()
-    log.info("start children update")
     updated_children = await taxonomy.update_node_children(entry, incoming_data)
-    log.info("end children update")
-    log.info(f"return children {updated_children}")
     return updated_children
 
 
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -265,16 +265,13 @@ class TaxonomyGraph:
             params["status"] = status
         query.extend(
             [
-                "RETURN n, size(error_node.errors) AS error_count",
+                "WITH n, size(error_node.errors) AS errors_count",
+                "RETURN n{.*, errors_count: errors_count}",
                 "ORDER BY n.created_at",
             ]
         )
         result = await get_current_transaction().run("\n".join(query), params)
-        # pack errors count in result
-        result = [
-            [dict(node, errors_count=num_errors)] for node, num_errors in await result.value()
-        ]
-        return result
+        return await async_list(result)
 
     async def add_node_to_end(self, label, entry):
         """
@@ -517,14 +514,17 @@ class TaxonomyGraph:
 
         # Stores result of last query executed
         result = []
-        for child in created_child_ids:
+        children_ids = created_child_ids + existing_ids
+        for child_id in children_ids:
             # Create new relationships if it doesn't exist
             query = f"""
                 MATCH (parent:{self.project_name}:ENTRY), (new_child:{self.project_name}:ENTRY)
-                WHERE parent.id = $id AND new_child.id = $child
+                WHERE parent.id = $id AND new_child.id = $child_id
                 MERGE (new_child)-[r:is_child_of]->(parent)
             """
-            _result = await get_current_transaction().run(query, {"id": entry, "child": child})
+            _result = await get_current_transaction().run(
+                query, {"id": entry, "child_id": child_id}
+            )
             result = list(await _result.value())
 
         return result
--- a/backend/editor/graph_db.py
+++ b/backend/editor/graph_db.py
@@ -20,11 +20,6 @@ txn.set(None)
 session = contextvars.ContextVar("session")
 session.set(None)
 
-test = contextvars.ContextVar("test")
-test.set(None)
-
-i = 0
-
 
 @contextlib.asynccontextmanager
 async def TransactionCtx():
@@ -32,15 +27,13 @@ async def TransactionCtx():
     Transaction context will set global transaction "txn" for the code in context.
     Transactions are automatically rollback if an exception occurs within the context.
     """
-    global txn, session, test, i
-    ival = i = i + 1
+    global txn, session
     try:
         async with driver.session() as _session:
             txn_manager = await _session.begin_transaction()
             async with txn_manager as _txn:
                 txn.set(_txn)
                 session.set(_session)
-                test.set(ival)
                 yield txn, session
     finally:
         txn.set(None)
--- a/parser/openfoodfacts_taxonomy_parser/parser.py
+++ b/parser/openfoodfacts_taxonomy_parser/parser.py
@@ -449,7 +449,7 @@ class Parser:
         """Create node to list parsing errors"""
         multi_label = self.create_multi_label(taxonomy_name, branch_name)
         query = f"""
-            CREATE (n:n:{multi_label}:ERRORS)
+            CREATE (n:{multi_label}:ERRORS)
             SET n.id = $project_name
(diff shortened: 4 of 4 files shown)


backend/editor/entries.py
```diff
@@ -483,29 +493,27 @@ def update_node_children(self, entry, new_children_ids):
...
             main_language_code = child.split(":", 1)[0]
-            created_node_id = self.create_node("ENTRY", child, main_language_code)
+            created_node_id = await self.create_node("ENTRY", child, main_language_code)
             created_child_ids.append(created_node_id)
 
             # TODO: We would prefer to add the node just after its parent entry
-            self.add_node_to_end("ENTRY", created_node_id)
+            await self.add_node_to_end("ENTRY", created_node_id)
 
         # Stores result of last query executed
```
aadarsh-ram, line 518: I think we forget to create relations if the nodes are already existing. So, the function from this line, needs to be changed as follows:
```
        children_to_create = created_child_ids + existing_ids
        # Stores result of last query executed
        result = []
        for child in children_to_create:
            # Create new relationships if it doesn't exist
            query = f"""
                MATCH (parent:{self.project_name}:ENTRY), (new_child:{self.project_name}:ENTRY)
                WHERE parent.id = $id AND new_child.id = $child
                MERGE (new_child)-[r:is_child_of]->(parent)
            """
            _result = await get_current_transaction().run(query, {"id": entry, "child": child})
            result = list(await _result.value())

        return result
```

backend/editor/api.py
```diff
@@ -149,8 +151,8 @@ async def list_all_projects(response: Response, status: Optional[StatusFilter] =
     List projects created in the Taxonomy Editor with a status filter
     """
     # Listing all projects doesn't require a taxonomy name or branch name
-    taxonony = TaxonomyGraph("", "")
-    result = list(taxonony.list_projects(status))
+    taxonomy = TaxonomyGraph("", "")
+    result = taxonomy.list_projects(status)
```
aadarsh-ram, line 155:
```suggestion
    result = await taxonomy.list_projects(status)
```

backend/editor/api.py
```diff
@@ -409,8 +400,10 @@ async def edit_entry(request: Request, branch: str, taxonomy_name: str, entry: s
     """
     taxonomy = TaxonomyGraph(branch, taxonomy_name)
     incoming_data = await request.json()
-    result = taxonomy.update_nodes("ENTRY", entry, incoming_data)
-    updated_entry = list(result)
+    log.info("start entry update")
+    updated_entry = await taxonomy.update_nodes("ENTRY", entry, incoming_data)
+    log.info("end entry update")
+    log.info(f"return {updated_entry}")
```
aadarsh-ram, line 406: Can we remove this log?

backend/editor/api.py
```diff
@@ -423,8 +416,10 @@ async def edit_entry_children(request: Request, branch: str, taxonomy_name: str,
     """
     taxonomy = TaxonomyGraph(branch, taxonomy_name)
     incoming_data = await request.json()
-    result = taxonomy.update_node_children(entry, incoming_data)
-    updated_children = list(result)
+    log.info("start children update")
+    updated_children = await taxonomy.update_node_children(entry, incoming_data)
+    log.info("end children update")
+    log.info(f"return children {updated_children}")
```
aadarsh-ram, line 422: Can we remove this too?

aadarsh-ram (changes requested):
Thanks for the amazing work! I requested some changes, where the code's not working due to the changes to async are breaking. Please do check them out.

backend/editor/api.py
alexgarel: yes totally ! I added it when I was debugging !

alexgarel, line 406:
```suggestion
    updated_entry = await taxonomy.update_nodes("ENTRY", entry, incoming_data)
```

backend/editor/api.py
alexgarel: yes again :-)

alexgarel, line 422:
```suggestion
    updated_children = await taxonomy.update_node_children(entry, incoming_data)
```

backend/editor/graph_db.py
```diff
@@ -3,34 +3,48 @@
...
+
+
 txn = contextvars.ContextVar("txn")
 txn.set(None)
 
 session = contextvars.ContextVar("session")
 session.set(None)
 
+test = contextvars.ContextVar("test")
+test.set(None)
```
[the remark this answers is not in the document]
alexgarel: Haha I though I had removed them :sweat_smile: 

I will do !

parser/openfoodfacts_taxonomy_parser/parser.py
```diff
@@ -369,7 +373,7 @@ def create_nodes(self, filename, multi_label):
 
     def create_previous_link(self, multi_label):
         self.parser_logger.info("Creating 'is_before' links")
-        query = f"MATCH(n:{multi_label}) WHERE exists(n.is_before) return n.id, n.is_before"
+        query = f"MATCH(n:{multi_label}) WHERE n.is_before IS NOT NULL return n.id, n.is_before"
```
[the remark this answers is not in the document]
alexgarel: @aadarsh-ram we need to store it temporarily in node, but we could remove it just after link creations (I though it was already the case). We can open a ticket for that, but it's not a big deal if it remains (but it makes a good first issue :-) )

backend/editor/entries.py
alexgarel: Cool we also fix bugs while reviewing :-)

backend/editor/graph_db.py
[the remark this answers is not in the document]
alexgarel: At some point as was not sure if our transaction mechanism was working as expected, that's why I added this… and it works very well :-)

parser/openfoodfacts_taxonomy_parser/parser.py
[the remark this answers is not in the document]
alexgarel: https://github.com/openfoodfacts/taxonomy-editor/issues/183

alexgarel:
@aadarsh-ram modifications done.

backend/editor/graph_db.py
```diff
@@ -3,34 +3,41 @@
...
+import logging
 
-from neo4j import GraphDatabase  # Interface with Neo4J
+import neo4j  # Interface with Neo4J
 
 from . import settings  # Neo4J settings
 from .exceptions import SessionMissingError  # Custom exceptions
 from .exceptions import TransactionMissingError
 
+log = logging.getLogger(__name__)
```
aadarsh-ram, line 14: I think we don't use this anywhere in this file, so I think this line and the imports can be removed

backend/requirements.txt
```diff
@@ -1,20 +1,20 @@
 anyio==3.6.1
 click==8.1.3
-fastapi==0.79.0
+fastapi==0.89.1
```
aadarsh-ram, line 3: @alexgarel Hopefully all the necessary dependencies for fastAPI and other packages has also been updated.

docker/dev.yml
```diff
@@ -20,6 +20,9 @@ services:
       # for linting / checks purpose
       - ./backend/pyproject.toml:/code/pyproject.toml
       - ./backend/setup.cfg:/code/setup.cfg
+    # uvicorn with live reload
+    command: ["uvicorn", "editor.api:app", "--host", "0.0.0.0", "--port", "80", "--reload", "--reload-dir", "/parser", "--reload-dir", "/code/editor"]
```
aadarsh-ram, line 24: Will be useful for dev! :)

aadarsh-ram (approved):
@alexgarel I'll approve. Feel free to merge after adding any necessary changes,

backend/editor/graph_db.py
alexgarel: Yep, but we should log more (at least debug or info) so I think we can keep it (it's a good pattern to have it)

backend/requirements.txt
alexgarel: seems so :-)

docker/dev.yml
alexgarel: Yes I though it was already done !

PR #185, opened 2023-01-19
feat: store languages marked as visible by user in localstorage

diivi:
Remember languages that the user selected as "Shown" in the UI with local storage.

### DIFF
--- a/taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
+++ b/taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
@@ -37,14 +37,13 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
 
   // Used for addition of a translation language
   const handleAddTranslation = (key) => {
-    const newRenderedTranslations = [
-      ...renderedTranslations,
-      { languageCode: key, tags: [] },
-    ];
-    setRenderedTranslations(newRenderedTranslations);
     key = "tags_" + key; // LC must have a prefix "tags_"
     const uuidKey = key + "_uuid"; // Format for the uuid
 
+    if (nodeObject[key]) {
+      // If the key already exists, do nothing
+      return;
+    }
     // Make changes to the parent NodeObject
     setNodeObject((prevState) => {
       const newNodeObject = { ...prevState };
@@ -70,11 +69,7 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
 
   const handleDialogConfirm = (newShownLanguageCodes) => {
     newShownLanguageCodes.forEach((languageCode) => {
-      if (
-        !renderedTranslations.some((item) => item.languageCode === languageCode)
-      ) {
-        handleAddTranslation(languageCode);
-      }
+      handleAddTranslation(languageCode);
     });
     setShownLanguages(newShownLanguageCodes);
     localStorage.setItem(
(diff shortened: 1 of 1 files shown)


taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -140,7 +144,26 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
         }
       }
     });
-    setShownLanguages(allLanguageCodes);
+    // get shown languages from local storage if it exists else use all languages
```
nobeeakon, line 147: let's move this part on it's own useEffect

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -140,7 +144,26 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
+    try {
+      const shownLanguages = JSON.parse(localStorage.getItem("shownLanguages"));
+      if (shownLanguages) {
+        setShownLanguages(shownLanguages);
+        shownLanguages.forEach((languageCode) => {
+          if (
+            !otherLangTags.some((item) => item.languageCode === languageCode)
+          ) {
+            handleAddTranslation(languageCode);
```
nobeeakon, line 156: as said in the previous PR, let's modify the `handleAddTranslation` so it only adds something when it doesn't exist , the current implementation is overriding

diivi: Done

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
diivi: I'll have to recalculate `allLanguageCodes` with the same `nodeObject` dependency

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -140,7 +139,25 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
         }
       }
     });
-    setShownLanguages(allLanguageCodes);
+    // get shown languages from local storage if it exists else use all languages
+    try {
+      const lsShownLanguages = JSON.parse(
+        localStorage.getItem("shownLanguages")
+      );
+      if (lsShownLanguages) {
```
nobeeakon, line 147: please rename to `localStorageShownLanguages`
We do need to check that the data it contains is correct: 1) if is an array, 2) if is an array of strings, 3) filter all the items that are valid language codes. Only after those validations we can use the data

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
nobeeakon: the shownLanguages shouldn't depend on the `nodeObject`, but on the languages stored in the local storage. Having it this way means that it will depend on the `nodeObject` so, for example,  every time we reload the site it will override the stored values, same thing will happen if the nodeObject changes... so not the behavior we want. This should only be a representation of the localStorage values and be in sync with it

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.jsx
```diff
@@ -70,13 +69,13 @@ const ListTranslations = ({ nodeObject, setNodeObject }) => {
...
-      if (
-        !renderedTranslations.some((item) => item.languageCode === languageCode)
-      ) {
-        handleAddTranslation(languageCode);
-      }
+      handleAddTranslation(languageCode);
     });
     setShownLanguages(newShownLanguageCodes);
+    localStorage.setItem(
+      "shownLanguages",
```
nobeeakon, line 76: this key should live outside of the component (be declared before the component...), `const SHOWN_LANGUAGES_KEY`

nobeeakon (approved):
nice!!

Issue #173, opened 2023-01-11
hide synonyms / stopwords in “root nodes” page


alexgarel:
We currently display synonyms and stopwords in "root nodes" page.

We shouldn't:

1. because change in synonyms and stopwords are rare, and must be done by advanced users (they have a lot of implications
2. we do not support editing them (and it's not a priority). If we support editing, in the future, they would have their own page.

So this issue is about not listing them in "root nodes" page.

Jagrutiti:
Does this issue some backend work too? Or is just FE?

diivi:
Alex mentioned this in out last meeting, it is backend only.

Jagrutiti:
> Alex mentioned this in out last meeting, it is backend only.

It is assigned to you so I was confused.

Jagrutiti:
Hey @alexgarel 

Here do I have to filter [get_all_root_nodes() ](https://github.com/openfoodfacts/taxonomy-editor/blob/main/backend/editor/entries.py#L350) function in `entries.py`?

I came to the conclusion because [the root node API](https://github.com/openfoodfacts/taxonomy-editor/blob/main/backend/editor/api.py#L180) calls that function.

Also for query, I should refer to [Neo4J ](https://neo4j.com/developer/cypher/filtering-show/)syntax?

diivi:
@Jagrutiti I have opened a PR for this issue as I though it was unassigned - #187

Jagrutiti:
> @Jagrutiti I have opened a PR for this issue as it was unassigned - #187

My bad.  Since you were working on the FE part I made the wrong assumption. I just saw your commits.

Issue #108, opened 2022-10-10
Support for addition of synonyms and stopwords in frontend


aadarsh-ram:
The frontend currently allows only addition of nodes of type "ENTRY". Addition of stopwords and synonyms through the frontend must be also supported and the POST requests for addition must be sent to the appropriate endpoint (`/nodes`) in the backend.

vedant-z:
Hey @aadarsh-ram I figured out that I need to add support for stopwords and synonyms in this part of code `<option value={'entry'}>Entry</option>` in `search/SearchResults.jsx` and `allentries/index.jsx` .

I need some help on what keywords should I add in option value or what needs to be done other than that?

Also in the backend part I need to make changes in backend/entries.py right?

aadarsh-ram:
Hey @vedant-z. Sorry for the late reply :(

You're right about the changes in frontend part, options have to be added. The frontend currently has a function to send a POST request to add an **entry** which is handled by `handleAddNode()`. But, synonyms/stopwords need a separate handler function to POST the ID as the string `synonym` or `stopword` and the main_language to the same endpoint (`newNode` variable is not required for addition of synonyms and stopwords).

The backend part (`entries.py`) requires a lot of work to include this change, since many functions have to be changed, namely `create_node()`. The `create_node()` function needs to be changed to include the suffix number by looking up the last stopword/synonym (Eg: stopwords:1 where 1 is the suffix number). 

Another function needs to be written to add this newly created node in the position next to the last stopwords/synonyms node in the Neo4j database. (The current `add_node_to_beginning()` function cannot be used, because order is important here). 

If you'd like to work on this issue, you'd require understanding of the way in which a taxonomy is parsed and stored in the database. As a start, I'd suggest you to read this [reference document](https://github.com/openfoodfacts/taxonomy-editor/blob/main/doc/references/taxonomy-parser.md) for the parsing process and test out the application in dev using the instructions given [here](https://github.com/openfoodfacts/taxonomy-editor/blob/main/doc/introduction/setup-dev.md). 

Sorry for the long comment, but hopefully this covers everything that needs to be changed in order to bring support for this. Feel free to ask any doubts!

aadarsh-ram:
@alexgarel Feel free to include anything that I might have missed :)

alexgarel:
@aadarsh-ram we should create a separate issue for backend part.

[...]

PR #187, opened 2023-01-20
feat: hide synonyms and stopwords in the root nodes page

diivi:
### What
- Exclude synonyms and stopwords from the `/rootnodes` endpoint as we don't support editing them anyway.

### Screenshot
image

### Fixes bug(s)
- #173 
- Invalidates - #108

### DIFF
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -352,7 +352,11 @@ class TaxonomyGraph:
         Helper function used for getting all root nodes in a taxonomy
         """
         query = f"""
-            MATCH (n:{self.project_name}) WHERE NOT (n)-[:is_child_of]->() RETURN n
+            MATCH (n:{self.project_name})
+            WHERE NOT (n)-[:is_child_of]->()
+            AND NOT (n: STOPWORDS)
+            AND NOT (n: SYNONYMS)
+            RETURN n
         """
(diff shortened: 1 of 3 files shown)


aadarsh-ram:
@alexgarel Do you think we need a separate endpoint (like `/rootentries`) for not listing stopwords and synonyms, or would the current `/rootnotes` be sufficient?

backend/editor/entries.py
```diff
@@ -352,7 +352,11 @@ def get_all_root_nodes(self):
         Helper function used for getting all root nodes in a taxonomy
         """
         query = f"""
-            MATCH (n:{self.project_name}) WHERE NOT (n)-[:is_child_of]->() RETURN n
+            MATCH (n:{self.project_name})
+            WHERE NOT (n)-[:is_child_of]->()
+            AND NOT (n: STOPWORDS)
+            AND NOT (n: SYNONYMS)
+            RETURN n
```
alexgarel, line 359: I'm not sure but what about 
```suggestion
            MATCH (n:{self.project_name}:ENTRY)
            WHERE NOT (n)-[:is_child_of]->()
            RETURN n
```
Do we need objects that are not entries ?

alexgarel (approved):
LGTM, just a comment, but I'm not sure, so I validate as is :-)

PR #189, opened 2023-01-20
build: avoid loosing build/.empty

### LINKED ISSUE #188: Avoid loosing taxonomy-editor-frontend/build/.empty every time we do a build (or run make frontend_quality)
### What
When a developper runs `make frontend_quality` the file `taxonomy-editor-frontend/build/.empty` is removed by the frontend build. 

This is a problem because we need to keep this file in git, so that the build/ folder exists and get the right ownership (otherwise, docker creates it on the fly but with root owning it, leading to an error on frontend build).

### Expected behavior

After a build, the folder `taxonomy-editor-frontend/build/` should still contains the `.empty` file (along with the generated assets).

I imagine it's possible to tell the build command to keep this .empty file as if it was a regular asset (we can add it in the src/ folder it's not a problem).

alexgarel:
when we run make frontend_quality

see:
- Fixes #188

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,8 @@ frontend_quality:
 	@echo "🍜 Quality checks JS"
 	${DOCKER_COMPOSE} run --rm taxonomy_node npx prettier -c src/
 	${DOCKER_COMPOSE} run --rm -e CI=true taxonomy_node npm run build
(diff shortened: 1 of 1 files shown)


aadarsh-ram (approved):
LGTM

PR #193, opened 2023-01-20
refactor: Typescript migration

nobeeakon:
### What
- refactored start project page into typescript. Replaced text used as labels, now using the labels Material UI provides (good for consistency across the site). 
- Also relocated utils file and migrated it to typescript. 

### Screenshot
imagen


taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -1,62 +1,62 @@
...
+  const [taxonomyName, setTaxonomyName] = useState("");
   const [description, setDescription] = useState("");
   const [loading, setLoading] = useState(false);
-  const [errorMessage, setErrorMessage] = useState(null);
+  const [errorMessage, setErrorMessage] = useState("");
   const navigate = useNavigate();
 
   const handleSubmit = () => {
-    const baseUrl = createBaseURL(taxonomyName, branchName);
+    if (!taxonomyName || !branchName) return;
```
diivi, line 32: Might be just a personal prefrence, but 
```suggestion
    if (!(taxonomyName && branchName)) return;
```

nobeeakon: I'd prefer to keep it as is (return if any of those two is missing)

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -68,69 +68,73 @@ const StartProject = () => {
...
+          variant="contained"
+          sx={{ mt: 3 }}
           onClick={handleSubmit}
-          sx={{ mt: 4, width: 130 }}
-          text="Submit"
-        ></LoadingButton>
+          disabled={!branchName || !taxonomyName || loading}
+        >
+          {!!loading ? <CircularProgress size={24} /> : "Submit"}
+        </Button>
```
aadarsh-ram, line 126: Since we are changing this, can we add text saying "Importing taxonomy.." with the circular progress? There is no indication to the user that the process is actually happening, except the circular progress.

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
```diff
@@ -6,7 +6,7 @@ import { API_URL } from "../../constants";
 import MaterialTable from "@material-table/core";
 import EditIcon from "@mui/icons-material/Edit";
 import { useEffect } from "react";
-import { toSnakeCase, toTitleCase } from "../../components/interConvertNames";
+import { toSnakeCase, toTitleCase } from "../../utils";
```
aadarsh-ram, line 9: Please make sure that the imports are updated everywhere.

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -68,69 +68,73 @@ const StartProject = () => {
...
-              else setTaxonomyName(null);
-            }}
-            renderInput={(params) => <TextField {...params} />}
-          ></Autocomplete>
-        </Stack>
-        <Stack direction="row" alignItems="center">
-          <Typography sx={{ mr: 8 }} variant="h5">
-            Branch Name
-          </Typography>
+        <div>
```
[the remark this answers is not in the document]
nobeeakon: yeah, think is ok to leave as <div> for the moment. `<Box>` is an abstracion of `div` and is useful when adding spacing and so on. Here is only a `div` and we're not using any of the additional stuff that `Box` provides

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -68,69 +68,73 @@ const StartProject = () => {
...
+              {TAXONOMY_NAMES.map((taxonomyItem) => (
+                <MenuItem value={taxonomyItem} key={taxonomyItem}>
+                  {taxonomyItem}
+                </MenuItem>
+              ))}
+            </Select>
+          </FormControl>
+        </div>
+
+        <div>
```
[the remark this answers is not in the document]
nobeeakon: yeah, think is ok to leave as <div> for the moment. <Box> is an abstracion of div and is useful when adding spacing and so on. Here is only a div and we're not using any of the additional stuff that Box provides

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -1,62 +1,62 @@
...
       method: "POST",
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify(dataToBeSent),
     })
       .then(async (response) => {
         const responseBody = await response.json();
-        if (!response.ok && responseBody.detail) {
-          throw new Error(responseBody.detail);
+        if (!response.ok) {
+          throw new Error(responseBody?.detail ?? "Unable to import");
```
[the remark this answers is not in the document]
nobeeakon: yeah, so, `!response.ok` could happen for a number of reasons we may not control, thus the `responseBody` is not guaranteed to exist so all this line is doing is adding a default value in case the `responseBody?.detail` doesn't exist

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -68,69 +68,73 @@ const StartProject = () => {
...
-          ></Autocomplete>
-        </Stack>
-        <Stack direction="row" alignItems="center">
-          <Typography sx={{ mr: 8 }} variant="h5">
-            Branch Name
-          </Typography>
+        <div>
+          <FormControl fullWidth sx={{ width: 265, mt: 4 }}>
+            <InputLabel id="taxonomy-name-label">Taxonomy</InputLabel>
+            <Select
```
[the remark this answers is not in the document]
nobeeakon: yeah, autocomplete is ok for cases when you want to allow the user to select from a list of options but allow them to use any other value that doesn't exist in the options. A good example is google search, when you are typing they give you a list of options, but you are not restricted to it. Here I don't think we want them to input any value they want, but to use one of the values we provide, thus I think the easiest way is to have a `select` with the list of options. Using an autocomplete here gives some additional complexity to the UX that I think don't give a real advantage vs the `select`, for example we need to validate the input, show error messages on invalid values, etc., which I think just makes it more cumbersome for the code and the user.

taxonomy-editor-frontend/src/pages/startproject/index.tsx
nobeeakon: done, added the text

taxonomy-editor-frontend/src/pages/gotoproject/index.jsx
nobeeakon: double checked this, and everything cool

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -68,69 +68,73 @@ const StartProject = () => {
...
+              labelId="taxonomy-name-label"
+              id="taxonomy-name"
+              value={taxonomyName}
+              label="Taxonomy"
+              onChange={(event) =>
+                setTaxonomyName(event.target.value as string)
```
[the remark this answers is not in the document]
nobeeakon: good question, not sure why I added it but now removed

aadarsh-ram (approved):
LGTM, thanks for the refactor!

PR #186, opened 2023-01-20
feat: parsing errors page

diivi:
* Added a page to show all errors related to a taxonomy
* Added the Errors column to /gotoproject table
image
image
image

FIxes #155

PR #197, opened 2023-01-23
refactor: navbar to typescript.

nobeeakon:
### What
- Summary: Navbar to typescript. Updated the way nav links are being set.

Changes:

- App.js is now App.tsx
- ResponsiveNavBar.jsx is now ResponsiveNavBar.tsx
- now instead of a useEffect that is called and sets the same navbar links in various places we have a function that does it and is passed to the places that need it
- Minor cleanup in various places (removed unused variables, etc.)
- sorted imports: first react, then react-router, then all the MUI ones, and last our own files…this will help us in the future (easier to read and clean), but as is just reordering I preferred to do it now..


taxonomy-editor-frontend/src/App.tsx
```diff
@@ -0,0 +1,94 @@
...
+      taxonomyName: string;
+      branchName: string;
+    }) => {
+      const urlPrefix = `${taxonomyName}/${branchName}/`;
+
+      const newNavLinks = [
+        { url: urlPrefix + "entry", translationKey: "Nodes" },
+        { url: urlPrefix + "search", translationKey: "Search" },
+        { url: urlPrefix + "export", translationKey: "Export" },
+      ];
```
diivi, line 46: Just a reminder to add the error page link once #186 gets merged

taxonomy-editor-frontend/src/App.tsx
```diff
@@ -0,0 +1,94 @@
...
+    },
+    []
+  );
+
+  return (
+    <ThemeProvider theme={theme}>
+      <CssBaseline />
+      <Router>
+        <ResponsiveAppBar displayedPages={navLinks} />
+        <div>
```
diivi, line 58: Is the `div` tag here just for syntactic sugar? I tried removing it and couldn't notice a difference

nobeeakon: done, removed the `div`

aadarsh-ram (approved):
LGTM @nobeeakon, thanks :)

PR #208, opened 2023-01-25
refactor: typescript, migrated gotoproject page

nobeeakon:
### What
- typescript: migrated 'Go to projects ' page to typescript

### Screenshot
same UI

### DIFF
--- a/taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
+++ b/taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
@@ -20,10 +20,10 @@ type ProjectType = {
 };
 
 type Props = {
-  resetNavLinks: () => void;
+  clearNavBarLinks: () => void;
 };
 
-const GoToProject = ({ resetNavLinks }: Props) => {
+const GoToProject = ({ clearNavBarLinks }: Props) => {
   const [projectData, setProjectData] = useState<ProjectType[]>([]);
   const navigate = useNavigate();
 
@@ -59,9 +59,9 @@ const GoToProject = ({ resetNavLinks }: Props) => {
 
   useEffect(
     function cleanMainNavLinks() {
-      resetNavLinks();
+      clearNavBarLinks();
     },
-    [resetNavLinks]
+    [clearNavBarLinks]
   );
 
   if (isError) {
--- a/taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
+++ b/taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
@@ -8,34 +8,54 @@ import EditIcon from "@mui/icons-material/Edit";
 import useFetch from "../../components/useFetch";
 import { API_URL } from "../../constants";
 import { toSnakeCase, toTitleCase } from "../../utils";
+import type { ProjectsAPIResponse } from "../../backend-types/types";
 
-const GotoProject = ({ resetNavLinks }) => {
-  const {
-    data: incomingData,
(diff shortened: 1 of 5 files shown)


taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
```diff
@@ -8,34 +8,54 @@ import EditIcon from "@mui/icons-material/Edit";
...
-          id: Math.random().toString(), // Used as Material Table component key
-          projectName: projectNode["id"],
-          taxonomyName: toTitleCase(projectNode["taxonomy_name"]),
-          branchName: projectNode["branch_name"],
-          description: projectNode["description"],
-          errors_count: projectNode["errors_count"],
+    let newProjects: ProjectType[] = [];
+
+    if (data) {
+      const backendProjects = data
```
aadarsh-ram, line 38: Why do we need two arrays `backendProjects` and `newProjects` for setting the state of `projectData`? It seems to be repetitive.

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
```diff
@@ -8,34 +8,54 @@ import EditIcon from "@mui/icons-material/Edit";
...
-  } = useFetch(`${API_URL}projects?status=OPEN`);
-  const [projectData, setProjectData] = useState([]);
+type ProjectType = {
+  id: string;
+  projectName: string;
+  taxonomyName: string;
+  branchName: string;
+  description: string;
+  errors_count: number;
+};
```
aadarsh-ram, line 20: Would it be better if we stored this in a separate `frontend-types/types.ts` folder?

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
```diff
@@ -8,34 +8,54 @@ import EditIcon from "@mui/icons-material/Edit";
...
+
+type Props = {
+  resetNavLinks: () => void;
+};
```
aadarsh-ram, line 24: `Props` seems to be too generic. A better name for resetting the navLinks would be helpful.

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
nobeeakon: I think is easier to follow when you set a variable in one go (after the `.map`), instead of defining it inside a `.forEach`. This way you know the complete set of information that is being defining. On the other side, `newProjects` are the new value that the react state will take and `backendProjects` are the projects the backend is returning.

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
nobeeakon: think is ok to keep it in here, it's better to keep things as nearest to the place they are used. In this case for the moment it will be used only here. For `backend-types` does make sense to have a single source of thruth for the types as those affect all the frontend and not only parts of it.

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
nobeeakon: Think is ok to name it `Props` as the information that the components receive is called props. In case we have multiple components in the same file, I'd agree that we shouldn't name it only `Props`. Renamed `resetNavLinks` to `clearNavBarLinks`

aadarsh-ram (approved):
Thanks @nobeeakon!

PR #216, opened 2023-02-03
refactor: API projects response

nobeeakon:
### What
- `/projects` API response is now a flat array . Before it was an array inside another array.
- updated the back and front end.

### Screenshot
same UI

### DIFF
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -287,7 +287,7 @@ class TaxonomyGraph:
 
         query_result = await get_current_transaction().run("\n".join(query), params)
 
-        return [item async for result_list in query_result for item in result_list ]
+        return [item async for result_list in query_result for item in result_list]
 
     async def add_node_to_end(self, label, entry):
         """
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -270,11 +270,13 @@ class TaxonomyGraph:
             "MATCH (n:PROJECT)",
             "OPTIONAL MATCH (error_node:ERRORS {branch_name: n.branch_name, id: n.id})",
         ]
+        
         params = {}
         if status is not None:
             # List only projects matching status
             query.append("WHERE n.status = $status")
             params["status"] = status
+            
         query.extend(
             [
                 "WITH n, size(error_node.errors) AS errors_count",
@@ -282,8 +284,16 @@ class TaxonomyGraph:
                 "ORDER BY n.created_at",
             ]
         )
-        result = await get_current_transaction().run("\n".join(query), params)
-        return await async_list(result)
+        
+        queryResult = await get_current_transaction().run("\n".join(query), params)
+        
+        result = []
+        
+        async for queryResultList in queryResult:
+            for item in queryResultList:
+                result.append(item)
+        
+        return result
 
     async def add_node_to_end(self, label, entry):
         """
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -285,15 +285,9 @@ class TaxonomyGraph:
             ]
         )
 
-        queryResult = await get_current_transaction().run("\n".join(query), params)
+        query_result = await get_current_transaction().run("\n".join(query), params)
 
-        result = []
-
-        async for queryResultList in queryResult:
-            for item in queryResultList:
-                result.append(item)
-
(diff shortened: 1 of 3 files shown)


backend/editor/entries.py
```diff
@@ -270,20 +270,30 @@ async def list_projects(self, status=None):
...
-        return await async_list(result)
+
+        queryResult = await get_current_transaction().run("\n".join(query), params)
+
+        result = []
+
+        async for queryResultList in queryResult:
+            for item in queryResultList:
+                result.append(item)
+
```
aadarsh-ram, line 295: @alexgarel @nobeeakon Do you think there's a better way to do this? I'd probably like to change the neo4j query to remove the nested lists.

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
```diff
@@ -35,21 +35,18 @@ const GoToProject = ({ clearNavBarLinks }: Props) => {
...
-        .filter((projectsArrayItem) => projectsArrayItem[0])
-        .map((projectsArrayItem) => {
-          const { id, branch_name, taxonomy_name, description, errors_count } =
-            projectsArrayItem[0];
-
+      const backendProjects = data.map(
+        ({ id, branch_name, taxonomy_name, description, errors_count }) => {
           return {
-            id: Math.random().toString(), // Used as Material Table component key
+            id, // needed by MaterialTable as key
```
aadarsh-ram, line 41: Just to be sure, you're using the projectName attribute given by the API, as id for Material Table?

backend/editor/entries.py
diivi: Couldn't figure out how to do this in Neo4j, but if we decide to do it in python, this can flatter the array:
  ```python
  result = await get_current_transaction().run(" ".join(query), params)
  return [record["n"] for record in await result.data()]
```

Comment (alexgarel) [reply backend/editor/entries.py:295]: It's not a big problem to unnest lists in the code. API readability is more important, it's quite fast to do this in python.

File: backend/editor/entries.py [hunk 3]
```diff
@@ -270,20 +270,30 @@ async def list_projects(self, status=None):
...
             [
                 "WITH n, size(error_node.errors) AS errors_count",
                 "RETURN n{.*, errors_count: errors_count}",
                 "ORDER BY n.created_at",
             ]
         )
-        result = await get_current_transaction().run("\n".join(query), params)
-        return await async_list(result)
+
+        queryResult = await get_current_transaction().run("\n".join(query), params)
```
alexgarel, line 288: Remember you are in the Python side, here you have to use snake_case, CamelCase is reserved to classes.

backend/editor/entries.py
```diff
@@ -270,20 +270,30 @@ async def list_projects(self, status=None):
...
+
+        result = []
+
+        async for queryResultList in queryResult:
+            for item in queryResultList:
+                result.append(item)
```
alexgarel, line 294: You can write it in a more pythonic way:

```suggestion
        result = [item async for result_list in query_result for item in result_list ]
```

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
alexgarel: @aadarsh-ram as every project is unique it seems ok, isn't it ? And it's more consistent.

alexgarel (approved):
LGTM generally but you have to change some variable case ! (yes it's always tricky when you do js and python at the same time to remember the conventions).

alexgarel:
@Jagrutiti beware that this PR might change your PR on models.py a bit.

backend/editor/entries.py
nobeeakon: done, I was trying to avoid the double list comprehension hehe, but yeah, let's go with python style

backend/editor/entries.py
nobeeakon: done, replaced with snake_case

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
nobeeakon: as far as I can see it is ok

PR #223, opened 2023-02-08
feat: setup fastapi tests

diivi:
### What
- Configure tests for FastAPI and build a base to write more tests later.
- docs - https://fastapi.tiangolo.com/tutorial/testing/

### DIFF
--- a/backend/editor/test_api.py
+++ b/backend/editor/test_api.py
@@ -1,0 +1,20 @@
+from datetime import datetime
+
+from fastapi.testclient import TestClient
+
+from .api import app
+
+client = TestClient(app)
+
+
+"""
+@app.get("/ping")
+async def pong(response: Response):
+    pong = datetime.now()
+    return {"ping": "pong @ %s" % pong}
+"""
+
+
+def test_ping():
+    response = client.get("/404")
+    assert response.status_code == 404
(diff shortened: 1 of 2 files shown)


backend/editor/test_api.py
```diff
@@ -0,0 +1,20 @@
+from datetime import datetime
+
+from fastapi.testclient import TestClient
+
+from .api import app
+
+client = TestClient(app)
```
alexgarel, line 7: you should not use a global variable but a fixture https://docs.pytest.org/en/latest/explanation/fixtures.html#about-fixtures

backend/Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN groupadd -g $USER_GID off && \
 
 COPY backend/requirements*.txt /code/
 COPY backend/editor /code/editor
+COPY backend/tests /code/tests
```
alexgarel, line 21: I would say we don't need test in the image (no need for them in prod), but we should bind mount them in docker-compose.yml instead (so you never have to rebuild)

alexgarel: I have pushed a commit to fix that.

alexgarel (approved):
Cool !

PR #225, opened 2023-02-09
Add tests

diivi:
### What
- Add some more tests for importing/uploading taxonomies


backend/tests/conftest.py
```diff
@@ -1,9 +1,23 @@
...
+
 import pytest
 from fastapi.testclient import TestClient
+from neo4j import GraphDatabase
 
 from editor.api import app
 
 
+# clean database before running tests
+@pytest.fixture(scope="session", autouse=True)
```
alexgarel, line 11: I would say we want it not at session level but for each test !

I would separate it in two part:

* a session fixture to get the driver
* a utility function to clean db. In most test files, where you use database, you could add a setup method, in autouse which calls this the clean db (and eventually setup some data).

backend/tests/data/test.txt
```diff
@@ -0,0 +1,37 @@
+# test taxonomy
```
alexgarel, line 1: FYI I have an updated version of test.txt in openfoodfacts-server, with some more complex DAG. https://github.com/openfoodfacts/openfoodfacts-server/blob/main/taxonomies/test.txt

backend/tests/test_api.py
```diff
@@ -2,3 +2,62 @@ def test_hello(client):
...
+    assert response.json().get("ping").startswith("pong @")
+
+
+def test_import_from_github(client):
+    response = client.post(
+        "/test/testing_branch/import",
+        json={"description": "test_description"},
+    )
+    assert response.status_code == 200
+    assert response.json() is True
```
alexgarel, line 19: do this test really use Github ?

I'm not keen on test depending on external servers, we should better mock the github functions (the one from the library).

backend/tests/test_api.py
```diff
@@ -2,3 +2,62 @@ def test_hello(client):
...
+            "/test_taxonomy/test_branch/upload",
+            files={"file": f},
+            data={"description": "test_description"},
+        )
+    assert response.status_code == 200
+    assert response.json() is True
+
+
+def test_add_taxonomy_duplicate_branch_name(client):
+    with open("tests/data/test.txt", "rb") as f:
```
alexgarel, line 34: you don't know the order in which test will be run, you have to import you taxonomy twice in this test. (and that's why we clean db at each test)

backend/tests/test_api.py
```diff
@@ -2,3 +2,62 @@ def test_hello(client):
...
+        response = client.post(
+            "/test_taxonomy/invalid-branch-name/upload",
+            files={"file": f},
+            data={"description": "test_description"},
+        )
+    assert response.status_code == 400
+    assert response.json() == {"detail": "branch_name: Enter a valid branch name!"}
+
+
+def test_add_taxonomy_duplicate_project_name(client):
```
alexgarel, line 55: same as for test above, your test must not depend on the last state of another test.

alexgarel (changes requested):
Good work, still you really have to have tests run in isolation.

alexgarel:
@diivi, FYI, @perierc will try to finish this PR :-)

alexgarel:
@perierc you don't need to use rebase, as normally we squash merge PRs at the end. So better use simple "merge" in the PR (easier to follow changes).

backend/tests/test_api.py
```diff
@@ -18,37 +19,101 @@ def test_hello(client):
...
+def test_ping(client):
+    response = client.get("/ping")
     assert response.status_code == 200
-    assert response.json() == {"message": "Deleted 1 projects"}
+    assert response.json().get("ping").startswith("pong @")
+
+
+def test_import_from_github(client, mocker):
+    mocker.patch("editor.api.TaxonomyGraph.is_branch_unique", return_value=True)
+    mocker.patch("editor.api.TaxonomyGraph.import_from_github", return_value=True)
```
perierc, line 30: @alexgarel please tell me what you think about how I mocked the calls to Github (I struggle to mock directly the GithubOperations class). 
Also, should I create the corresponding integration tests which use the real Github servers?

alexgarel: As seen together:

* we will patch github.Github instead
* replace all `from github import Github` by `import github` and `Github` by `github.Github` so that mocking once affect all use of it.

perierc:
I've added a test for the loading and dumping scripts and made them work in Docker.
However the scripts work for the `test-neo4j.json` file that seems to have an outdated structure, but the dump does not work with the `test` taxonomy that can be imported from the UI.
I have to update the scripts to make them match the new taxonomy structure.

alexgarel (approved):
Great !

Issue #219, opened 2023-02-07
Refactor: remove the need of tag_ids_str


alexgarel:
We changed Neo4J version.

With this version the search engine can handle arrays better than before.
This might alleviate the need of a specific field for the search (like tags_ids_string).

Try to remove those specific fields.

PR #228, opened 2023-02-14
refactor: Removed tags_str property

aadarsh-ram:
### What
- Removes the tags_str property
- Uses the tags_lc property for creating search indexes

### Related issue(s)
- #219


alexgarel (approved):
Great !

Surprising that you had no tests to change !

Issue #168, opened 2023-01-10
create project: feedback if you use hyphens in branch name


alexgarel:
Adding hyphens and special characters is not possible in branch nam, due to limitations of Neo4J labels. 
We require a feedback mechanism for indicating users about the name in the frontend.

NihalRadhakrishna:
@alexgarel  can I work on this. I was thinking of showing error message like this
![image](https://user-images.githubusercontent.com/91626278/218311434-e9a1bab8-e96f-455d-bcb4-f766a616f0b7.png)

alexgarel:
Yes, please do @NihalRadhakrishna.
Thank you for showing an example.

PR #229, opened 2023-02-14
feat: Show feedback when special characters are entered in branch name

NihalRadhakrishna:
### What
- Show feedback when special characters are entered in branch name.
- If there is error, disabled the submit button.

### Screenshot
image

### Fixes bug(s)
- #168  

### Part of 
- #30

### DIFF
--- a/taxonomy-editor-frontend/src/pages/startproject/index.tsx
+++ b/taxonomy-editor-frontend/src/pages/startproject/index.tsx
@@ -135,7 +135,7 @@ const StartProject = ({ clearNavBarLinks }) => {
           variant="contained"
           sx={{ mt: 3 }}
           onClick={handleSubmit}
-          disabled={!branchName || !taxonomyName || loading}
+          disabled={!branchName || !taxonomyName || loading || showBranchNameError}
         >
           {loading ? (
             <>
--- a/taxonomy-editor-frontend/src/pages/startproject/index.tsx
+++ b/taxonomy-editor-frontend/src/pages/startproject/index.tsx
@@ -98,12 +98,12 @@ const StartProject = ({ clearNavBarLinks }) => {
         <div>
           <TextField
             error={showBranchNameError}
-            helperText={showBranchNameError && "Special characters or white spaces are not allowed"}
+            helperText={showBranchNameError && "Special characters, capital letters and white spaces are not allowed"}
             size="small"
             sx={{ width: 265, mt: 2 }}
             onChange={(event) => {
               setBranchName(event.target.value);
-              const specialChar = /[!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?\s]+/;
+              const specialChar = /[!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?\sA-Z]+/;
               if(specialChar.test(event.target.value)){
                 setShowBranchNameError(true);
               }
--- a/taxonomy-editor-frontend/src/pages/startproject/index.tsx
+++ b/taxonomy-editor-frontend/src/pages/startproject/index.tsx
@@ -26,7 +26,7 @@ const StartProject = ({ clearNavBarLinks }) => {
   const [description, setDescription] = useState("");
   const [loading, setLoading] = useState(false);
   const [errorMessage, setErrorMessage] = useState("");
-  const [branchNameError, setBranchNameError] = useState(false);
+  const [showBranchNameError, setShowBranchNameError] = useState(false);
   const navigate = useNavigate();
 
   useEffect(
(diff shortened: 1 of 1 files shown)


taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -96,10 +97,19 @@ const StartProject = ({ clearNavBarLinks }) => {
 
         <div>
           <TextField
+            error={branchNameError}
+            helperText={branchNameError && "Special characters or white spaces are not allowed"}
```
Jagrutiti: As far as I remember even upper case characters are not allowed in branch name. 

You should test and confirm though.

nobeeakon: does this means we restrict to a-z, 0-9 and _ ?, if so, maybe we could test for that, think it would be easier that way

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -26,6 +26,7 @@ const StartProject = ({ clearNavBarLinks }) => {
   const [description, setDescription] = useState("");
   const [loading, setLoading] = useState(false);
   const [errorMessage, setErrorMessage] = useState("");
+  const [showBranchNameError, setShowBranchNameError] = useState(false);
```
nobeeakon, line 29: no need for this to be a state, it could be 
`const validBranchName  = RegexpValidation.test(branchName)`
and used in the `error` and `helperText` props

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -96,10 +97,19 @@ const StartProject = ({ clearNavBarLinks }) => {
 
         <div>
           <TextField
+            error={showBranchNameError}
+            helperText={showBranchNameError && "Special characters or white spaces are not allowed"}
             size="small"
             sx={{ width: 265, mt: 2 }}
             onChange={(event) => {
               setBranchName(event.target.value);
+              const specialChar = /[!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?\s]+/;
```
nobeeakon, line 106: please check the other comment,  on the state and the regexp

taxonomy-editor-frontend/src/pages/startproject/index.tsx
NihalRadhakrishna: @Jagrutiti yeah, I tested , the caps is also not allowed, so added check for that ! thanks.

taxonomy-editor-frontend/src/pages/startproject/index.tsx
NihalRadhakrishna: Yes @nobeeakon I will change that! Thanks.

NihalRadhakrishna:
@nobeeakon I have removed the extra state and changed the regex as u suggested. Thanks!

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -64,6 +65,9 @@ const StartProject = ({ clearNavBarLinks }) => {
     setErrorMessage("");
   };
 
+
+  const textFieldError = BranchNameRegEx.test(branchName);
```
nobeeakon, line 69: think this will be `true` if is the branchName is valid.  Please rename as `isValidBranchName`

NihalRadhakrishna: @nobeeakon I have changed the name. Thanks!

nobeeakon: looks great, just please update the places where this variable is used, like disabled and helper text, and error, to `!isValidBranchName`

NihalRadhakrishna: Actually the above statement returns true if there exists a special character(invalid name). It returns false if branch name is valid. So should I just change the name to `isInvalidBranchName`

nobeeakon: yes please use `isInvalidBranchName`, also please use `branchNameRegEx` instead of `BranchNameRegEx`  . Sorry for the confusion.

NihalRadhakrishna:
I have made the necessary changes. Thanks!

nobeeakon (approved):
Great work!! 
Please run prettier and update the PR [so we can merge](https://github.com/openfoodfacts/taxonomy-editor/actions/runs/4178565151/jobs/7237630153#step:3:674)

NihalRadhakrishna:
@nobeeakon I ran prettier.  Thanks!

Issue #227, opened 2023-02-11
feat: Use load animation


NihalRadhakrishna:
### Problem
When user clicks on open existing project button or edit project button it shows "loading..." text.

### Proposed solution
We can add loading animation using mui circular progress component.

### Mockups
![image](https://user-images.githubusercontent.com/91626278/218259987-2f95374c-4d4b-435c-adac-fdd0c6389d2c.png)

### Part of
- #

NihalRadhakrishna:
Something like this...
![image](https://user-images.githubusercontent.com/91626278/218260265-128ae0e3-d6fc-43bd-a620-d6c1f955beed.png)

aashimawadhwa:
I would like to take this issue.

alexgarel:
@aashimawadhwa just tell you gonna take it and it's yours :-) (if no one already signal he is on it).

NihalRadhakrishna:
@aashimawadhwa can I pls take this issue since I opened it. I've already got it working on my local setup.

PR #232, opened 2023-02-16
feat: add loading animation

NihalRadhakrishna:
### What
- created Loader component
- where ever there is loading text replaced with Loader component.

### Fixes bug(s)
- #227 

### Part of 
- #30


taxonomy-editor-frontend/src/components/Loader.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { Box } from "@mui/material";
+import CircularProgress from "@mui/material/CircularProgress";
+
+const Loader = () => {
+  return (
+    <Box
+      sx={{
+        width: "100%",
```
nobeeakon, line 9: is this width needed? think by default it is a div which has width: 100%

taxonomy-editor-frontend/src/components/Loader.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
```
nobeeakon, line 1: I think it is ok to replace the 'loading...' message with a spinner. But for the moment I'd say we could just use the `CircularProgress` directly, without the need for a wrapper, this gives us more control on the spacing, size, etc. Please instead of importing this component, use the `CircularProgress`.

taxonomy-editor-frontend/src/components/Loader.tsx
nobeeakon, line 9: is this `width:100%` needed? from what I could see it is a `div` by default

taxonomy-editor-frontend/src/components/Loader.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Box } from "@mui/material";
+import CircularProgress from "@mui/material/CircularProgress";
+
+const Loader = () => {
+  return (
+    <Box
+      sx={{
+        width: "100%",
+        textAlign: "center",
+        py: 10,
```
nobeeakon, line 11: please when replacing the `Loader` component the margins to the places where used, for example, having `py:10` feels like a lot of space in the children or parent list

taxonomy-editor-frontend/src/components/Loader.tsx
```diff
@@ -0,0 +1,20 @@
...
+        m: 0,
```
nobeeakon, line 12: is it needed to override the margins to 0?

NihalRadhakrishna:
Sorry for the unnecessary css! I have made the changes you suggested. Also reduced margin value in child components.

aadarsh-ram:
@NihalRadhakrishna I think the Errors page loader, has still not be changed. Please do check that out. 
There seems to be two loader components for the "entry" page. Please remove the loader wherever necessary.

taxonomy-editor-frontend/src/pages/editentry/ListEntryChildren.jsx
```diff
@@ -87,9 +88,14 @@ const ListEntryChildren = ({ url, urlPrefix, setUpdateNodeChildren }) => {
   }
   if (isPending) {
```
Piv94165, line 89: @nobeeakon I am taking this PR
I can see the 2 loading spins  when we want to edit a node.
Do we want a circular progress when we are waiting for children and parents ? Or do we only need a circular progress for the whole page ?
I ask these questions to know if I delete these circular progress components in ListEntryChildren and ListEntryParents. If we want to keep them, we will display both of them or do we only want one for both lists ?

alexgarel: @Piv94165 let's keep it simple ! Spinners are not the main UX blocker right now ;-)

alexgarel:
I close, this was resolved here : https://github.com/openfoodfacts/taxonomy-editor/pull/232

alexgarel:
BTW thanks @NihalRadhakrishna  for participating :-)

PR #237, opened 2023-04-14
api: Delete taxonomy project

### LINKED ISSUE #139: User should be able to close project without exporting it
### What
A user might want to abandon a project, and we should give him the ability to close it.

It does not remove the project from the database, it just changes the status.

It would be also nice to be able to re-open a closed project.

Jagrutiti:
### What
- Added an API to delete taxonomy projects

### Closes
#139

### DIFF
--- a/backend/editor/api.py
+++ b/backend/editor/api.py
@@ -513,3 +513,12 @@ async def delete_node(request: Request, branch: str, taxonomy_name: str):
     incoming_data = await request.json()
     id = incoming_data["id"]
     await taxonomy.delete_node(taxonomy.get_label(id), id)
+
+
+@app.delete("/{taxonomy_name}/{branch}/delete")
+async def delete_project(response: Response, branch: str, taxonomy_name: str):
+    """
+    Delete a project
+    """
+    taxonomy = TaxonomyGraph(branch, taxonomy_name)
+    await taxonomy.delete_taxonomy_project(branch, taxonomy_name)
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -608,3 +608,15 @@ class TaxonomyGraph:
         _result = await get_current_transaction().run(query, params)
         result = [record["node"] for record in await _result.data()]
         return result
+
+    async def delete_taxonomy_project(self, branch, taxonomy_name):
+        """
+        Delete taxonomy projects
+        """
+
+        query = f"""
+            MATCH (n:{self.project_name}) DETACH DELETE n
+        """
+
+        result = await get_current_transaction().run(query)
+        return result


backend/tests/test_api.py
```diff
@@ -1,4 +1,42 @@
+import pytest
+from backend.editor.api import api
+from falcon import testing
+
+
+@pytest.fixture(autouse=True)
+def test_setup(neo4j):
+    # delete all the nodes and relations in the database
+    query = "MATCH (n:p_test_branch:t_test:b_branch) DETACH DELETE n"
```
alexgarel, line 9: Let's really delete all :-)

```suggestion
    # delete all the nodes and relations in the database
    query = "MATCH n DETACH DELETE n"
```

backend/tests/test_api.py
```diff
@@ -1,4 +1,42 @@
...
+    return testing.TestClient(api)
+
 def test_hello(client):
     response = client.get("/")
     assert response.status_code == 200
     assert response.json() == {"message": "Hello user! Tip: open /docs or /redoc for documentation"}
+
+def delete_project_test(neo4j):
+    session = neo4j.session()
+    query = "MATCH (n:p_test_branch:t_test:b_branch:ENTRY) RETURN COUNT(*)"
```
aadarsh-ram, line 27: This query isn't being used anywhere. Please remove it.

backend/tests/test_api.py
```diff
@@ -1,4 +1,42 @@
...
+            CREATE (n:PROJECT)
+            SET n.id = test_project
+            SET n.taxonomy_name = testtaxonomyname
+            SET n.branch_name = testbranchname
+            SET n.description = testdescription
+            SET n.status = OPEN
+            SET n.created_at = datetime()
+        """
+    result = session.run(create_project)
+    import pdb; pdb.set_trace();
```
aadarsh-ram, line 39: Please try to keep imports at the top. There might be issues during linting.

diivi: I don't think this is supposed to be here either; maybe they used it for debugging and forgot to remove it.

Jagrutiti: I was running the test case when encountered errors. 

I committed the code to share what I was doing before posting the question on Slack. 

Apologies for any confusion.

backend/tests/test_api.py
```diff
@@ -1,42 +1,43 @@
...
+        SET n.created_at = datetime()
+    """
     result = session.run(create_project)
-    import pdb; pdb.set_trace();
-    stimulate_api_result = client.simulate_get("/testtaxonomyname/testbranchname/delete")
-    assert stimulate_api_result.status_code == 200
-    assert stimulate_api_result.json == {}
+    response = client.delete("/test_taxonomy_name/test_branch_name/delete")
+    assert response.status_code == 200
+    # assert response.json() == {"message": ""}
```
Jagrutiti, line 35: Though this test case passes, it does not completed the test of the API.

After invoking the delete API I do not get the count of number of nodes or relations delete. I tried this in Neo4j browser as well.

The query is straightforward. Which makes me suspect that my data is incomplete here. 

What nodes should I create for this test to be successful?

backend/tests/test_api.py
Jagrutiti: Done

backend/editor/entries.py
```diff
@@ -608,3 +608,14 @@ async def full_text_search(self, text):
...
         result = [record["node"] for record in await _result.data()]
         return result
+
+    async def delete_taxonomy_project(self, branch, taxonomy_name):
+        """
+        Delete taxonomy projects
+        """
+
+        query = f"""
+            MATCH (n:{self.project_name}) DETACH DELETE n
```
alexgarel, line 618: should'nt you add something like `RETURN count(n)` to have a return value ?

backend/editor/api.py
```diff
@@ -513,3 +513,12 @@ async def delete_node(request: Request, branch: str, taxonomy_name: str):
...
     await taxonomy.delete_node(taxonomy.get_label(id), id)
+
+
+@app.delete("/{taxonomy_name}/{branch}/delete")
+async def delete_project(response: Response, branch: str, taxonomy_name: str):
+    """
+    Delete a project
+    """
+    taxonomy = TaxonomyGraph(branch, taxonomy_name)
+    await taxonomy.delete_taxonomy_project(branch, taxonomy_name)
```
alexgarel, line 524: you don't return anything here.

Maybe try to return the number of removed node (if possible) and at least a message. (do not return directly the result of Neo4J, but make it more readable).

backend/tests/test_api.py
alexgarel: you didn't return anything in you method, that's why !

backend/editor/entries.py
Jagrutiti: Done

backend/tests/test_api.py
```diff
@@ -27,17 +27,29 @@ def test_delete_project(neo4j, client):
...
-    assert number_of_nodes == 0
+    record = result.single()
 
+    assert record['n.id'] == 'test_project'
+    assert record['n.taxonomy_name'] == 'test_taxonomy_name'
+    assert record['n.branch_name'] == 'test_branch_name'
+    assert record['n.description'] == 'test_description'
+    assert record['n.status'] == 'OPEN'
+    assert record['n.created_at'] is not None
+    assert record['n.project_name'] == "p_test_taxonomy_name_test_branch_name"
```
Jagrutiti, line 49: Hey @alexgarel 

Thank you for reviewing.

I have updated the test case. 

How do I make sure that the project with correct name is deleted? My message fails. 

This assert should succeed: `assert response.json() == {"message": "Deleted 1 projects"}`

But it says zero projects deleted.

alexgarel: Sorry @Jagrutiti I don't know what to say, I must test it myself, and those days I must concentrate on infrastructure.

backend/editor/entries.py
```diff
@@ -614,9 +614,13 @@ async def delete_taxonomy_project(self, branch, taxonomy_name):
...
 
-        query = f"""
-            MATCH (n:{self.project_name}) DETACH DELETE n RETURN count(n)
+        delete_query = """
+            MATCH (n:PROJECT {taxonomy_name: $taxonomy_name, branch: $branch})
+            DELETE n
         """
-        result = await get_current_transaction().run(query)
-        count = await result.single()
-        return count[0]
```
Jagrutiti: Hey @alexgarel 

I figured out what the issue was. It was in query and also in creating dummy data.

I have resolved it.

backend/tests/test_api.py
```diff
@@ -24,32 +24,31 @@ def test_delete_project(neo4j, client):
...
-    assert record['n.status'] == 'OPEN'
-    assert record['n.created_at'] is not None
-    assert record['n.project_name'] == "p_test_taxonomy_name_test_branch_name"
-    # check ends here
-
-    response = client.delete("/test_taxonomy_name/test_branch_name/delete")
+    session.run(create_project)
+    session.run(create_project2)
+
+    response = client.delete("/test_taxonomy_name/test_branch/delete")
```
Jagrutiti, line 47: This test case tests the API. 

Do you think I should also test the delete function in `entries.py`?

aadarsh-ram: I don't think it's required at this point, since this indirectly tests the database as well. @alexgarel what do you think?

alexgarel: I agree with @aadarsh-ram

backend/editor/entries.py
alexgarel: kudos

alexgarel (approved):
Kudos @Jagrutiti !

PR #277, opened 2023-11-02
refactor(backend): better error handling in backend

Piv94165:
Add specific messages if Github settings were not added and change some status codes in backend

### What
Some errors raised by backend have a 500 status code intead of having 4XX error.
For example, if a developer forgets to add a Personal Access Token and wants to import a project to edit, the server throws a 500 error instead of 400 error.

### Screenshot
image

### Part of 
- backend/editor

### DIFF
--- a/backend/editor/api.py
+++ b/backend/editor/api.py
@@ -138,6 +138,14 @@ async def validation_exception_handler(request: Request, exc: RequestValidationE
         content=jsonable_encoder({"detail": "Invalid request", "errors": reformatted_errors}),
     )
 
+@app.exception_handler(HTTPException)
+async def http_exception_handler(request: Request, exc: HTTPException):
+    """
+    Custom exception handler to log FastAPI exceptions.
+    """
+    # Log the detail message
+    log.info(f" ERROR: {exc.detail}")
+    return JSONResponse(status_code=exc.status_code, content={"message": exc.detail})
 
 # Get methods
 
--- a/backend/editor/api.py
+++ b/backend/editor/api.py
@@ -24,6 +24,7 @@ from fastapi import (
     status,
 )
 from fastapi.encoders import jsonable_encoder
+from fastapi.exception_handlers import http_exception_handler
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import FileResponse, JSONResponse
@@ -111,7 +112,7 @@ def file_cleanup(filepath):
     try:
         os.remove(filepath)
     except FileNotFoundError:
-             log.warn(f"Taxonomy file {filepath} not found for deletion")
+        log.warn(f"Taxonomy file {filepath} not found for deletion")
 
 
 class StatusFilter(str, Enum):
@@ -140,14 +141,13 @@ async def validation_exception_handler(request: Request, exc: RequestValidationE
 
 
 @app.exception_handler(HTTPException)
-async def http_exception_handler(request: Request, exc: HTTPException):
+async def log_http_exception(request: Request, exc: HTTPException):
     """
     Custom exception handler to log FastAPI exceptions.
     """
     # Log the detail message
     log.info(f" ERROR: {exc.detail}")
-    return JSONResponse(status_code=exc.status_code, content={"message": exc.detail})
-    # return await http_exception_handler(request, exc)
+    return await http_exception_handler(request, exc)
 
 
 # Get methods
--- a/backend/editor/api.py
+++ b/backend/editor/api.py
@@ -138,6 +138,7 @@ async def validation_exception_handler(request: Request, exc: RequestValidationE
         content=jsonable_encoder({"detail": "Invalid request", "errors": reformatted_errors}),
     )
 
+
 @app.exception_handler(HTTPException)
 async def http_exception_handler(request: Request, exc: HTTPException):
     """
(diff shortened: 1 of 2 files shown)


backend/editor/api.py
```diff
@@ -101,7 +101,7 @@ def check_single(id):
     if len(id) == 0:
         raise HTTPException(status_code=404, detail="Entry not found")
     elif len(id) > 1:
-        raise HTTPException(status_code=500, detail="Multiple entries found")
+        raise HTTPException(status_code=409, detail="Multiple entries found")
```
eric-nguyen-cs, line 104: I believe that this should still be a 500 error: this error is not user-based and should not happen (if it happens, it means that we have not properly checked the nodes we have added)
```suggestion
        raise HTTPException(status_code=500, detail="Multiple entries found")
```

backend/editor/api.py
```diff
@@ -111,7 +111,7 @@ def file_cleanup(filepath):
     try:
         os.remove(filepath)
     except Exception as e:
-        raise HTTPException(status_code=500, detail="Taxonomy file not found for deletion") from e
+        raise HTTPException(status_code=404, detail="Taxonomy file not found for deletion") from e
```
eric-nguyen-cs, line 114: Not even sure if a HTTPException should be raised: 
- file_cleanup is used in a FastAPI background task so it ran after the HTTP response
- if `os.remove` raises the `FileNotFoundError` error, then the file is already cleaned up. I believe that logging the unexpected behaviour should be enough  

```suggestion
        except FileNotFoundError:
             log.warn(f"Taxonomy file {filepath} not found for deletion")  
        
```

backend/editor/api.py
```diff
@@ -139,6 +139,16 @@ async def validation_exception_handler(request: Request, exc: RequestValidationE
...
 
 
+@app.exception_handler(HTTPException)
+async def http_exception_handler(request: Request, exc: HTTPException):
+    """
+    Custom exception handler to log FastAPI exceptions.
+    """
+    # Log the detail message
+    log.info(f" ERROR: {exc.detail}")
+    return JSONResponse(status_code=exc.status_code, content={"message": exc.detail})
```
eric-nguyen-cs, line 149: https://fastapi.tiangolo.com/tutorial/handling-errors/#override-the-httpexception-error-handler

I believe we could reuse the default `http_exception_handler`

```suggestion
    return await http_exception_handler(request, exc)
```

backend/editor/github_functions.py
```diff
@@ -47,19 +59,28 @@ def update_file(self, filename):
...
+            )
+        except GithubException as e:
+            # Handle GitHub API-related exceptions
+            raise Exception(f"GitHub API error: {e}")
+        except FileNotFoundError:
+            # Handle file not found error (e.g., when 'filename' does not exist)
+            raise Exception(f"File not found: {filename}")
+        except Exception as e:
+            # Handle any other unexpected exceptions
+            raise Exception(f"An error occurred: {e}")
```
eric-nguyen-cs, line 83: nit: To keep the exception context

```suggestion
        except GithubException as e:
            # Handle GitHub API-related exceptions
            raise Exception(f"GitHub API error: {e}") from e
        except FileNotFoundError:
            # Handle file not found error (e.g., when 'filename' does not exist)
            raise Exception(f"File not found: {filename}") from e
        except Exception as e:
            # Handle any other unexpected exceptions
            raise Exception(f"An error occurred: {e}") from e
```

eric-nguyen-cs (commented):
Some few nits but otherwise LGTM

backend/editor/api.py
alexgarel: I agree it's ok to just log.

backend/editor/github_functions.py
alexgarel: good catch.

backend/editor/api.py
Piv94165: I'm not sure to understand what I should do, because returning  http_exception_handler creates a loop since I am overriding the HTTPException handler

alexgarel: the precise link to doc is https://fastapi.tiangolo.com/tutorial/handling-errors/#re-use-fastapis-exception-handlers

So add a:
```python
import fastapi.exception_handlers
```

and here:
```python
    return await fastapi.exception_handlers.http_exception_handler(request, exc)
```

alexgarel: Oh I see you resolved it indeed…

backend/editor/github_functions.py
```diff
@@ -21,8 +22,19 @@ def init_driver_and_repo(self):
...
         Initalize connection to Github with an access token
         """
-        github_driver = Github(settings.access_token)
-        repo = github_driver.get_repo(settings.repo_uri)
+        access_token = settings.access_token
+        if not access_token:
+            raise HTTPException(
+                status_code=400,
+                detail="Access token is not set. Please add your access token in .env",
+            )
```
alexgarel, line 30: is the access_token mandatory for all operations ?

I think it's only needed for the create pull request. I so, I would move this check to the create pull request method.

Piv94165: It is also necessary to import a project. If we haven't specified it in the .env file, we can't effectively test the app as there is no data

alexgarel: ok !

alexgarel (approved):
LGTM

Issue #261, opened 2023-10-19
Using Poetry for Python dependency management


eric-nguyen-cs:
Right now, we are using pip and venv to manage Python dependencies, which can be quite error-prone

Poetry has become a well established tool to manage Python dependencies on larger projects
https://python-poetry.org
```[tasklist]
### Tasks
- [ ] Adding to backend poetry: add poetry, dependencies, update dockerfile for dependency installation, update makefile and docs
```

eric-nguyen-cs:
https://github.com/openfoodfacts/robotoff

PR #278, opened 2023-11-04
chore: Add Poetry for Python dependency management

eric-nguyen-cs:
### What
Adding Poetry for Python dependency management:
- Now using Poetry for Python dependency management
- Update the relevant Dockerfile to now use poetry.lock instead of requirements.txt to install dependencies
- Revamp documentation for local development

### Fixes bug(s)
https://github.com/openfoodfacts/taxonomy-editor/issues/261


backend/pyproject.toml
```diff
@@ -32,4 +36,30 @@ ensure_newline_before_comments = true
 line_length = 100
 
 [tool.mypy]
-ignore_missing_imports = true
\ No newline at end of file
+ignore_missing_imports = true
+
+[tool.poetry]
+name = "backend"
```
alexgarel, line 42: openfoodfacts-taxonomy-editor seems better to me :-) ? Or should it map the folder name (in which case this is ok).
```suggestion
name = "openfoodfacts-taxonomy-editor"
```

docker/dev.yml
```diff
@@ -34,7 +47,7 @@ services:
         USER_UID: ${USER_UID:-1000}
         USER_GID: ${USER_GID:-1000}
     environment:
-      - REACT_APP_API_URL=//api.${TAXONOMY_EDITOR_DOMAIN}:${TAXONOMY_EDITOR_PORT:-80}/
+      - REACT_APP_API_URL=http://api.${TAXONOMY_EDITOR_DOMAIN}${PUBLIC_TAXONOMY_EDITOR_PORT:-80}/
```
alexgarel, line 50: don't add the http because in production we use https.

If you use "://" it uses the current page scheme.

taxonomy-editor-frontend/src/constants.js
```diff
@@ -1,21 +1,4 @@
...
-    components[0] = "api";
-    return location.protocol + "//" + components.join(".") + "/";
-  } else {
-    // this is a default for simple dev setup
-    return "http://localhost:80/";
-  }
-}
-
-export const API_URL = taxonomyApiUrlFromUi(window.location);
+export const API_URL = process.env.REACT_APP_API_URL;
```
alexgarel, line 1: No you can't do that, in production the file is static and asking process.env does not work (you are in the user browser), that's why it was done like it was !

alexgarel (changes requested):
Kudos for this query and the work done.

I just have a blocker.

docker/dev.yml
```diff
@@ -47,7 +47,7 @@ services:
         USER_UID: ${USER_UID:-1000}
         USER_GID: ${USER_GID:-1000}
     environment:
-      - REACT_APP_API_URL=http://api.${TAXONOMY_EDITOR_DOMAIN}${PUBLIC_TAXONOMY_EDITOR_PORT:-80}/
+      - REACT_APP_API_URL=://api.${TAXONOMY_EDITOR_DOMAIN}${PUBLIC_TAXONOMY_EDITOR_PORT:-80}/
```
alexgarel, line 50: I don't think you should have the ":" here !
```suggestion
      - REACT_APP_API_URL=//api.${TAXONOMY_EDITOR_DOMAIN}${PUBLIC_TAXONOMY_EDITOR_PORT:-80}/
```

alexgarel (approved):
I validate so you can merge it, but I think you have one correction to do on a settings.

Great work.

PR #284, opened 2023-11-09
fix(taxonomy-editor-frontend): redirect user to node editor when clicking from nodes page

Piv94165:
### What
User can not choose a node to edit from Nodes page.
To do that, has to go to Search page and click on it. 

### Screenshot
Go to https://ui.taxonomy.openfoodfacts.net/states/editstostates/entry
And try to click on an edit button (nothing happens)

### Fixes bug(s)
Now, on Nodes page, user can click on an edit button to go to node editor.

### Part of 
taxonomy-editor-frontend


Piv94165:
This change fixes the bug.
But it could be the opportunity to refactor a bit, since code is duplicated (search pages and nodes pages could use the same component)

alexgarel (approved):
Let's merge. Refactor is needed indeed, but I will open a specific issue.

PR #290, opened 2023-11-16
refactor(taxonomy-editor-frontend): Create a component for nodes table

Piv94165:
Create a component for nodes table used in nodes and search pages

### What
- Right now we have quite the same list on root nodes page and search page, we should use the same component to avoid double maintenance.

### Screenshot
no visible change

### Code pointers
taxonomy-editor-frontend\src\pages\search\SearchResults.tsx
taxonomy-editor-frontend\src\pages\root-nodes\index.tsx

### DIFF
--- a/taxonomy-editor-frontend/src/components/NodesTableBody.tsx
+++ b/taxonomy-editor-frontend/src/components/NodesTableBody.tsx
@@ -1,0 +1,42 @@
+import {
+  IconButton,
+  TableBody,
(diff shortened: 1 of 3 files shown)


alexgarel:
@nobeeakon would you eventually have the time to review this PR ?

alexgarel (approved):
Codes seems really ok and I tested it, it works.

Thanks @Piv94165 !

PR #304, opened 2023-12-07
Refactor/make parsing background task

Piv94165:
### What
- 

### Fixes bug(s)
- 

### Part of 
-


backend/editor/api.py
```diff
@@ -387,8 +389,8 @@ async def import_from_github(request: Request, branch: str, taxonomy_name: str):
     if not await taxonomy.is_branch_unique():
         raise HTTPException(status_code=409, detail="branch_name: Branch name should be unique!")
 
-    result = await taxonomy.import_from_github(description)
-    return result
+    background_tasks.add_task(await taxonomy.import_from_github(description))
```
eric-nguyen-cs, line 392:
```suggestion
    background_tasks.add_task(taxonomy.import_from_github, description)
```

Piv94165: Thanks @eric-nguyen-cs. But with 
`background_tasks.add_task(taxonomy.import_from_github, description)`
it seems that the task is never really done, It seems to send only an empty list to the frontend whereas with 
`background_tasks.add_task(await taxonomy.import_from_github(description))`
I can see all the logs from taxonomy-editor-taxonomy_api while the taxonomy is parsed.
So to me it seems to be more complex.

alexgarel: @Piv94165 
`background_tasks.add_task(await taxonomy.import_from_github(description))` calls the function but does not execute it in background (in Python we resolve arguments first)

backend/editor/entries.py
```diff
@@ -87,7 +87,8 @@ async def parse_taxonomy(self, filename):
         Helper function to call the Open Food Facts Python Taxonomy Parser
         """
         # Close current transaction to use the session variable in parser
-        await get_current_transaction().commit()
+        async with get_current_transaction() as transaction:
+            transaction.commit()
```
alexgarel, line 91: hum, it seems strange to me that you have to do this ! (but it could be, just I don't have the whole picture in my head now...)

Maybe put a `# TODO: verify if really needed after parser refactor`

Piv94165: you're right, while working on handling background task on frontend, I realize I shouldn't do this.

backend/editor/entries.py
```diff
@@ -82,59 +84,76 @@ async def create_node(self, label, entry, main_language_code):
...
-                # Parse taxonomy with given file name and branch name
-                parser_object(filename, self.branch_name, self.taxonomy_name)
-                return True
-            except Exception as e:
-                raise TaxonomyParsingError() from e
+        if uploadfile is None:  # taxonomy is imported
+            base_url = (
+                "https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-server"
+                "/main/taxonomies/"
+            )
```
alexgarel, line 95: you must use the REPO_URI env variable for that
```suggestion
            default_url = "https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-server"
            base_url = os.environ.get("REPO_URI", default_url) + "/main/taxonomies/"
```

backend/editor/entries.py
```diff
@@ -82,59 +84,76 @@ async def create_node(self, label, entry, main_language_code):
...
-            async with TransactionCtx():
-                await self.create_project(description)  # Creates a "project node" in neo4j
+        # Create a new transaction context and create a "project node" in Neo4j
+        async with TransactionCtx():
+            await self.create_project(description)
 
-            return status
-        except Exception as e:
-            raise TaxonomyImportError() from e
+        return True
```
alexgarel, line 139: You should do it before closing transaction and launching background task. It's more logical.

Also I don't think you need to close the transaction, as the session used  by parse_taxonomy is completely different.

Just create the project and launch the task.

```python
    await self.create_project(description)
    background_tasks.add_task(self.parse_taxonomy)
    return True
```

backend/editor/entries.py
```diff
@@ -82,59 +84,76 @@ async def create_node(self, label, entry, main_language_code):
...
-            status = await self.parse_taxonomy(filepath)
+            # Add the task to background tasks
+            background_tasks.add_task(self.parse_taxonomy, uploadfile)
+            print("Background task added")
             async with TransactionCtx():
                 await self.create_project(description)
-            return status
+            return True
         except Exception as e:
             raise TaxonomyImportError() from e
```
alexgarel, line 158: Same as above, just create project and launch task.
Also no exception there (exception, if any, will happen in the background task)
```python
    await self.create_project(description)
    background_tasks.add_task(self.parse_taxonomy, uploadfile)
    return True
```

backend/editor/entries.py
```diff
@@ -82,59 +84,76 @@ async def create_node(self, label, entry, main_language_code):
...
+                    # Create parser object and pass current session to it
+                    parser_object = parser.Parser(session)
+                    try:
+                        # Parse taxonomy with given file name and branch name
+                        parser_object(filepath, self.branch_name, self.taxonomy_name)
+                        return True
+                    except Exception as e:
+                        raise TaxonomyParsingError() from e
+        except Exception as e:
+            raise TaxonomyImportError() from e
```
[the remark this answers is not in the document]
Piv94165: You're right, if an exception is raised during parsing, no errors node is created. So I've added this node in this case.
Now if I want to change the project status in the background task, I find some difficulties to get current transaction. Indeed, in the background task I loose the connection with the db (I have no current transaction).
Do you have an idea how I could handle this ?

taxonomy-editor-frontend/src/pages/root-nodes/index.tsx
```diff
@@ -84,7 +85,7 @@ const RootNodes = ({
     );
   }
 
-  if (isPending || !nodes) {
+  if (isPending || !nodes || nodes.length === 0) {
```
perierc, line 88: I think there is an issue with this implementation: the message will show every time we open a project until the nodes are fetched, even if the project is already parsed and open, which is not what indicates the displayed message.

This message should be displayed only if the project status is LOADING

taxonomy-editor-frontend/src/components/Alerts.tsx
perierc, line 1: I believe the file should have the same name as the exported component 🤔

perierc (commented):
Arrived a bit late, but I think some comments can be interesting to consider 😉

taxonomy-editor-frontend/src/components/Alerts.tsx
```diff
@@ -0,0 +1,67 @@
...
+      {message}
+    </Alert>
+  );
+};
+
+// warning users the taxonomy had parsing errors, so should not edit it
+export const WarningParsingErrors: React.FC<WarningParsingErrorsProps> = ({
+  baseUrl,
+}) => {
+  console.log("in WarningParsingErrors");
```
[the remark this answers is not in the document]
Piv94165: oops you're right

PR #311, opened 2023-12-12
ci: fix grafana-annotation-action to avoid node deprecation

alexgarel:
see https://github.com/frankie567/grafana-annotation-action/issues/19


aadarsh-ram (approved):
LGTM

PR #316, opened 2023-12-20
refactor(backend,taxonomy-editor-frontend): Handle background task on frontend

Piv94165:
### Current situation
- If parsing takes too long, the frontend times out, resulting in a timeout notification : 

Image

- Only successfully parsed taxonomies are displayed in the existing projects list. Failed or loading taxonomies are not visible.

### To be done 

- [x] Project Creation Page: Users should land on a loading page upon creating a project, prompting them to refresh the page for updates
- [x] Projects Listing: Ensure all projects are visible in the table, including those that failed or are still loading
- [x]  ¨Project Status: Display project statuses (successfully parsed, parsing failed, parsing in progress) in the projects list

These enhancements will offer a more comprehensive view of project statuses, provide clarity to users during parsing, and ensure visibility of all projects in the list, even if parsing fails or takes longer than expected.

### Screenshot
image
A small text reminds user to reload the page to see taxonomy

image
In the list, the project status is displayed.
Not only the open projects are listed but also the other ones.

### Part of 
(https://github.com/orgs/openfoodfacts/projects/103/views/1?pane=issue&itemId=46878339)

### DIFF
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -87,7 +87,8 @@ class TaxonomyGraph:
         Helper function to call the Open Food Facts Python Taxonomy Parser
         """
         # Close current transaction to use the session variable in parser
-        await get_current_transaction().commit()
+        async with get_current_transaction() as transaction:
+            transaction.commit()
 
         with SyncTransactionCtx() as session:
             # Create parser object and pass current session to it
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -86,10 +86,6 @@ class TaxonomyGraph:
         """
         Helper function to call the Open Food Facts Python Taxonomy Parser
         """
-        # Close current transaction to use the session variable in parser
-        async with get_current_transaction() as transaction:
-            transaction.commit()
-
         with SyncTransactionCtx() as session:
             # Create parser object and pass current session to it
             parser_object = parser.Parser(session)
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -5,6 +5,7 @@ import re
 import tempfile
 import urllib.request  # Sending requests
 
+from fastapi import BackgroundTasks
 from openfoodfacts_taxonomy_parser import normalizer  # Normalizing tags
 from openfoodfacts_taxonomy_parser import parser  # Parser for taxonomies
 from openfoodfacts_taxonomy_parser import unparser  # Unparser for taxonomies
@@ -22,7 +23,6 @@ from .graph_db import (  # Neo4J transactions context managers
     TransactionCtx,
     get_current_transaction,
 )
-from fastapi import BackgroundTasks
 
 
 async def async_list(async_iterable):
@@ -91,7 +91,7 @@ class TaxonomyGraph:
             "https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-server"
             "/main/taxonomies/"
         )
-        filename = self.taxonomy_name + ".txt"
+        filename = f"{self.taxonomy_name}.txt"
         base_url += filename
         try:
             with tempfile.TemporaryDirectory(prefix="taxonomy-") as tmpdir:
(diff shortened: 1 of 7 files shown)


Piv94165:
Link: https://github.com/openfoodfacts/taxonomy-editor/pull/304#issue-2031007807 needs to be merged prior to this PR for correct functionality.

Remarks:
During the import of a taxonomy, the task runs in the background. However, the backend is unable to send any response due to the synchronous nature of the following functions:
```
self.create_nodes(filename, multi_label)
self.create_child_link(multi_label)
self.create_previous_link(multi_label)
self.create_fulltext_index(taxonomy_name, branch_name)
self.create_parsing_errors_node(taxonomy_name, branch_name)
```
This results in the page loading indefinitely if the user tries to access existing projects while a new project is being created, as the backend cannot respond to the frontend until parsing is complete.

It appears that the issue lies in the synchronous execution of these functions. Converting them to asynchronous operations could potentially resolve this problem.
The reason why the backend can't respond to frontend seems to be the synchronism of the functions above. They should be asynchronous I think.

backend/editor/entries.py
```diff
@@ -120,10 +120,12 @@ async def import_from_github(self, description):
                 status = await self.parse_taxonomy(filepath)  # Parse the taxonomy
 
             async with TransactionCtx():
-                await self.create_project(description)  # Creates a "project node" in neo4j
+                await self.set_project_status(status="OPEN")
 
             return status
         except Exception as e:
+            async with TransactionCtx():
+                await self.set_project_status(status="FAILED")
```
alexgarel, line 128: Maybe you could set project status FAILED only if create_project did not fail ? (create project could return created project and you could test it here, remember to create the variable with `None` before the `try` though to be sure it exists.

alexgarel (approved):
Cool. Small improvement proposed.

backend/editor/entries.py
```diff
@@ -82,59 +85,86 @@ async def create_node(self, label, entry, main_language_code):
...
         try:
             with tempfile.TemporaryDirectory(prefix="taxonomy-") as tmpdir:
                 # File to save the downloaded taxonomy
                 filepath = f"{tmpdir}/{filename}"
+                if uploadfile is None:
+                    # Downloads and creates taxonomy file in current working directory
+                    urllib.request.urlretrieve(base_url, filepath)
+                else:
+                    with open(filepath, "wb") as f:
+                        shutil.copyfileobj(uploadfile.file, f)
```
alexgarel, line 111: I think this part should be isolated in a function to make code more readable.

The function should be a context manager so it handles the tempfile transparently

```python

    @contextlib.contextmanager
    def get_taxonomy_file(self, uploadfile=None):
        ...
        with tempfile.TemporaryDirectory(prefix="taxonomy-") as tmpdir:
           ...
           yield filepath

...

    def parse_taxonomy(self, uploadfile=None):
        with self.get_taxonomy_file(uploadfile) as filepath:
            ...
```

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
```diff
@@ -29,22 +29,30 @@ const GoToProject = ({ clearNavBarLinks }: Props) => {
   const navigate = useNavigate();
 
   const { data, isPending, isError } = useFetch<ProjectsAPIResponse>(
-    `${API_URL}projects?status=OPEN`
+    `${API_URL}projects`
```
alexgarel, line 32: Maybe, in another PR I think we will have to provide a way to filter by status !

backend/editor/entries.py
```diff
@@ -82,59 +85,86 @@ async def create_node(self, label, entry, main_language_code):
...
+
+                with SyncTransactionCtx() as session:
+                    # Create parser object and pass current session to it
+                    parser_object = parser.Parser(session)
+                    try:
+                        # Parse taxonomy with given file name and branch name
+                        parser_object(filepath, self.branch_name, self.taxonomy_name)
```
alexgarel, line 118: In case of success, shouldn't you update the project status to "OPEN" ?

alexgarel (changes requested):
Thank @Piv94165 !

There are still a bit of things to improve.

alexgarel:
@Piv94165 should we close this PR ? If I remember well you merged those changes in another PR right ?

Issue #289, opened 2023-11-16
Make parsing more performant


eric-nguyen-cs:
## Now

Currently, it takes a few seconds to import some taxonomies and some taxonomies are not parsed before the http requests timeouts.
The parsing step takes too much time

## Goal

Understand why parsing takes so much time to respond and resolve it
- Make the parser faster if possible 
- If not possible, respond to the client asynchronously

alexgarel:
@eric-nguyen-cs it's mandatory to make the client respond asynchronously, (there are huge taxonomies). see #206

But working on parsing performance is also important as I believe there might be low hanging fruit.

PR #317, opened 2023-12-21
feat: Decouple parsing the taxonomy and writing the taxonomy to the database

eric-nguyen-cs:
### What
- The goal of this PR is to decouple the action of parsing the taxonomy `.txt` file into Python objects and the action of writing this data into the Neo4J database

### Note
This PR is quite big, so to simplify the review, you can go through commit by commit (I tried to make the commits quite small and atomic)

### Fixes bug(s)
https://github.com/openfoodfacts/taxonomy-editor/issues/88

### Part of 
- This work is the precursor of the parsing performance work. See [next PR](https://github.com/openfoodfacts/taxonomy-editor/issues/289)


Piv94165 (approved):
Indeed the clarity in this version is a significant improvement !

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -0,0 +1,199 @@
...
+            "errors_list": self.parser_logger.parsing_errors,
+        }
+        self.session.run(query, params)
+
+    def __call__(self, filename: str, branch_name: str, taxonomy_name: str):
+        """Process the file"""
+        branch_name = normalizing(branch_name, char="_")
+        multi_label = self._create_multi_label(taxonomy_name, branch_name)
+        taxonomy_parser = TaxonomyParser()
+        taxonomy = taxonomy_parser.parse_file(filename, self.parser_logger)
```
alexgarel, line 177: I would have made this a bit more composable.

For me there is the taxonomy parser that returns a Taxonomy,this is your TaxonomyParser and the taxonomy backend that takes a taxonomy and put it in Neo4J (so this class would have been Neo4jTaxonomy for example or Neo4jBackend.

Then a function (or callable class) would weaved them together. this function would be those four lines + a cll to the backend (to run the lines below). It would take the TaxonomyParser object and the Taxonomy Backend object as parameters.
It would make it clearer that you could add a PostgreSQL backend.

alexgarel (approved):
Kudos !

I have a proposal to change the composition pattern a bit, but it's all good.

parser/openfoodfacts_taxonomy_parser/parser/parser.py
eric-nguyen-cs: It seems a bit over-engineered considering that we are currently only writing to a Neo4J database
Do you think it is essential to do this now? If we ever add support for another database, we could do the refactor at that time

alexgarel: I'm ok to postpone it, though I was seeing this as a simple refactor that adds clarity.

Do as you wish :-)

Issue #183, opened 2023-01-19
Parser: remove is_before property after link creation


alexgarel:
As we create the graph in the parser, we temporarily store a is_before attribute. At the end of the process, based on this, we create the is_before link.
But we should also remove the is_before attribute just after, to make thinks more clear.

Cf. Parser.create_previous_link

mahak24342:
i wanna take up this issue , please assign me this issue

PR #318, opened 2023-12-21
feat: Improve parser performance

eric-nguyen-cs:
## What
The goal of this PR is to improve the performance of the writing of the taxonomy data to the `Neo4J` database, so that every taxonomy is parsable in a raisonnable amount of time

## Note

This PR is quite big, so to simplify the review, you can go through commit by commit (I tried to make the commits quite small and atomic)

## Changes
We can have small performance increases for small taxonomies, but the changes mainly target medium or large taxonomies, by:
- batching requests, as to not pay the cost of the query optimisation multiple times
- using range indexes on the node id property to drastically improve look up times in big taxonomies when creating relationships

## Fixes bug(s)
https://github.com/openfoodfacts/taxonomy-editor/issues/289
https://github.com/openfoodfacts/taxonomy-editor/issues/183

## Part of 
Follow up of this [parser/db writer decoupling PR](https://github.com/openfoodfacts/taxonomy-editor/pull/317)

## Results
We have an overall speedup of **x11** (mainly **730.6s -> 25.44s** for the `categories` taxonomy). Before, the writes were very slow, so HTTP timeouts were possible when importing a taxonomy file, but now all taxonomies are parsable in less than 30s.

### Details
Before:
Additives -> 34.93s
Allergens -> 3.31s
Amino Acids -> 1.6s
Categories -> **730.6s**
Countries -> **103.49s**
Data Quality -> 16.08s
Food Groups -> 1.63s
Improvements -> 1.06s
Ingredients -> **255.13s**
Ingredients Analysis -> 1.12s
Ingredients Processing -> 5.94s
Labels -> 56.08s
Languages -> 32.26s
Minerals -> 6.58s
Misc -> 0.64s
Nova Groups -> 0.61s
Nucleotides -> 1.64s
Nutrients -> 10.79s
Origins -> 12.40s
Other Nutritional Substances -> 2.84s
Packaging Materials -> 6.90s
Packaging Recycling -> 0.94s
Packaging Shapes -> 3.81s
Periods After Opening -> 1.34s
Preservation -> 0.57s
States -> 2.19s
Test ->  1.41s
Vitamins -> 2.94s
(Total: 1299s)

After:
All taxonomies are parsable in less than 30s
Additives -> 4.78s
Allergens -> 1.23s
Amino Acids -> 0.77s
Categories -> **25.44s**
Countries -> **15.71s**
Data Quality -> 1.09s 
Food Groups -> 0.96s
Improvements -> 1.41s
Ingredients -> **13.76s**
Ingredients Analysis -> 1.25s
Ingredients Processing -> 2.28s
Labels -> 3.95s
Languages -> 14.26s
Minerals -> 4.0s
Misc -> 0.32s
Nova Groups -> 0.71s
Nucleotides -> 1.95s
Nutrients -> 5.87s
Origins -> 5.41s
Other Nutritional Substances -> 1.5s
Packaging Materials -> 2.4s
Packaging Recycling -> 0.66s
Packaging Shapes -> 2.38s
Periods After Opening -> 0.63s
Preservation -> 0.57s
States -> 0.83s
Test -> 0.82s
Vitamins -> 2.40s
(Total: 117s)


eric-nguyen-cs:
I believe that the tests are not passing because the parent order written to the exported taxonomy file is not deterministic
Maybe solving [this issue](https://github.com/openfoodfacts/taxonomy-editor/issues/319) first is necessary to merge this PR and [the previous one](https://github.com/openfoodfacts/taxonomy-editor/pull/317)

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
+                if not key in seen_properties:
+                    additional_query += " SET n." + key + " = entry_node." + key + "\n"
+                    seen_properties.add(key)
+            for key in entry_node.tags:
+                if not key in seen_tags:
+                    additional_query += " SET n." + key + " = entry_node." + key + "\n"
+                    seen_tags.add(key)
+
+        query = base_query + additional_query
+        self.session.run(query, entry_nodes=[entry_node.to_dict() for entry_node in entry_nodes])
```
alexgarel, line 85: I'm quite positively surprised by the flexibility of Neo4j to run this even if there are properties is missing in some nodes :-)

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
+          UNWIND entry_nodes as entry_node
+          CREATE (n:{project_label}:ENTRY)
+          SET n.id = entry_node.id
+          SET n.preceding_lines = entry_node.preceding_lines
+          SET n.src_position = entry_node.src_position
+          SET n.main_language = entry_node.main_language
+        """
+        additional_query = ""
+        seen_properties = set()
+        seen_tags = set()
```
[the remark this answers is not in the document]
alexgarel: BTW as tags and properties seems to be set the same way, is it wise to separate in two set ? Or should we instead only have one ?

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
+                    seen_tags.add(key)
+
+        query = base_query + additional_query
+        self.session.run(query, entry_nodes=[entry_node.to_dict() for entry_node in entry_nodes])
+
+        self.parser_logger.info(
+            f"Created {len(entry_nodes)} ENTRY nodes in {timeit.default_timer() - start_time} seconds"
+        )
+
+    def _create_other_nodes(self, other_nodes: list[NodeData], project_label: str):
```
alexgarel, line 91: nit: putting it just after _create_other_node would make code easier to read.

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
-                    before_id,
-                )
-            elif not relation[0]:
-                self.parser_logger.error("link not created between %s and %s", id, before_id)
-
-    def create_child_link(self, child_links: list[ChildLink], multi_label: str):
-        """Create the relations between nodes"""
+        start_time = timeit.default_timer()
+
+        query = f"""
```
alexgarel, line 109: can you comment a bit on the query.

For example the importance of USING INDEX.

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
+        count = result.value()[0]
+
+        self.parser_logger.info(
+            f"Created {count} 'is_before' links in {timeit.default_timer() - start_time} seconds"
+        )
+
+        if count != len(previous_links):
+            self.parser_logger.error(
+                f"Created {count} 'is_before' links, {len(previous_links)} links expected"
+            )
```
alexgarel, line 130: I'm asking myself if it's not a show stopper. This is a very important part to be able to unparse the taxonomy. So it would be a pity we let a user edit a taxonomy that won't be rendered as a PR in the end !

We should at least add it to the error nodes of the taxonomy.

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
+    def _create_child_links(
+        self, child_links: list[ChildLink], entry_nodes: list[NodeData], project_label: str
+    ):
+        """Create the 'is_child_of' relations between nodes"""
         self.parser_logger.info("Creating 'is_child_of' links")
+        start_time = timeit.default_timer()
+
+        node_ids = set([node.id for node in entry_nodes])
+        normalised_child_links = []
+        unnormalised_child_links = []
```
alexgarel, line 141: You should comments chunk of codes, to guide the reader… your future self will thanks you for that :wink: 

```suggestion
        # we collect nodes with a parent id which is the id of an entry
        # and nodes where a synonym was used to designate the parent (unnormalised)
        normalised_child_links = []
        unnormalised_child_links = []
```

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
-            f"""CREATE FULLTEXT INDEX {project_name+'_SearchIds'} IF NOT EXISTS
-            FOR (n:{project_name}) ON EACH [n.id]\n"""
-            + """
-            OPTIONS {indexConfig: {`fulltext.analyzer`: 'keyword'}}"""
+            if child_link["parent_id"] in node_ids:
+                normalised_child_links.append(child_link)
+            else:
+                unnormalised_child_links.append(child_link)
+
+        normalised_query = f"""
```
alexgarel, line 148:
```suggestion
        # adding normalised links is easy as we can directly match parent entries
        normalised_query = f"""
```

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
+
+        language_codes = set()
+        for child_link in unnormalised_child_links:
+            lc, parent_id = child_link["parent_id"].split(":")
+            language_codes.add(lc)
+            child_link["parent_id"] = parent_id
+
+        parent_id_query = " OR ".join(
+            [f"child_link.parent_id IN p.tags_ids_{lc}" for lc in language_codes]
         )
```
alexgarel, line 168: This way you de-corelate the information between language and id, it seems dangerous to me, it's not sure that you won't match the wrong parent ! More over it's not the best for performances.

I would either run a query per language or change child_link to have language and id separated so that you can write: `child_link.parent_id IN p.tags_ids_{child_link.lc}` (this can be done by changing to dict())

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -1,158 +1,213 @@
...
+            count += unnormalised_result.value()[0]
+
+        self.parser_logger.info(
+            f"Created {count} 'is_child_of' links in {timeit.default_timer() - start_time} seconds"
+        )
+
+        if count != len(child_links):
+            self.parser_logger.error(
+                f"Created {count} 'is_child_of' links, {len(child_links)} links expected"
+            )
```
alexgarel, line 202: This is not a show stopper but we could add it to error nodes.

alexgarel (changes requested):
Really good work. 

I have some proposal though.

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -257,9 +250,7 @@ def _harvest_entries(self, filename: str, entries_start_line: int) -> Iterator[N
                         lc, value = self._get_lc_value(line[10:])
                     except ValueError:
                         self.parser_logger.error(
-                            "Missing language code at line %d ? '%s'",
-                            line_number + 1,
-                            ellipsis(line),
+                            f"Missing language code at line {line_number + 1} ? '{self.parser_logger.ellipsis(line)}'"
```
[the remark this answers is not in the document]
eric-nguyen-cs: Yes, we should probably improve the error handling (add some useful debugging info)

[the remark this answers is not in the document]
eric-nguyen-cs: Tracking this here: https://github.com/openfoodfacts/taxonomy-editor/issues/331

parser/openfoodfacts_taxonomy_parser/parser/parser.py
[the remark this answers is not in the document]
eric-nguyen-cs: I don't think it really matters: both give the same result, although I find slightly clearer to separate the two, as they store different types.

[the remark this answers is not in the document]
eric-nguyen-cs: Yes we could simplify the code here

parser/openfoodfacts_taxonomy_parser/parser/parser.py
eric-nguyen-cs: This is already added to the error node of the taxonomy.
I agree : errors in general should be fixed before modifying the taxonomy. The frontend should better reflect that (https://github.com/openfoodfacts/taxonomy-editor/issues/332)

parser/openfoodfacts_taxonomy_parser/parser/parser.py
eric-nguyen-cs: This is already the case (`parser_logger.error` append errors to a list which we then add to the error nodes)

parser/openfoodfacts_taxonomy_parser/parser/parser.py
eric-nguyen-cs: Yes thanks for finding the bug!

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -157,27 +162,31 @@ def _create_child_links(
             RETURN COUNT(relation)
         """
 
-        language_codes = set()
+        # for unnormalised links, we need to group them by language code of the parent id
+        lc_child_links_map = {}
```
alexgarel, line 166: nit: just FYI, this would be a good use case for a [collections.defaultdict](https://docs.python.org/3/library/collections.html#defaultdict-objects):
`lc_child_links_map = collections.defaultdict(list)`, no need to check for the existance of lc in map then :-)

parser/openfoodfacts_taxonomy_parser/parser/parser.py
alexgarel: oups, that's true ;-)

alexgarel (approved):
Great !

PR #336, opened 2024-01-17
fix(test): remove determinism of parent order in tests

eric-nguyen-cs:
### What
Currently, parent order is not deterministic when parsing and unparsing a taxonomy. Thus, the tests sometimes fail or pass (see https://github.com/openfoodfacts/taxonomy-editor/actions/runs/7555310972 and https://github.com/openfoodfacts/taxonomy-editor/actions/runs/7555383123/job/20570246361 that come from exactly the same code)

This is a short term fix: the goal is to add parent order determinism in a future PR (as well as good comment locality)

### DIFF
--- a/parser/tests/integration/test_parse_unparse_integration.py
+++ b/parser/tests/integration/test_parse_unparse_integration.py
@@ -50,24 +50,9 @@ def test_round_trip(neo4j):
         test_dumper = unparser.WriteTaxonomy(session)
         lines = list(test_dumper.iter_lines("p_test_branch"))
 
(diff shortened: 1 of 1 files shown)


eric-nguyen-cs:
Another short term fix would be to get rid of the two parents in the test file. Not sure what is preferable.
In any case, when we add the parent order determinism in the unparsing code, we will be able to add that test back

eric-nguyen-cs:
Instead of updating the tests to get the tests passing, I updated the code here https://github.com/openfoodfacts/taxonomy-editor/pull/337

PR #337, opened 2024-01-17
feat: Add parent order determinism

eric-nguyen-cs:
### What
Currently, when we export back the taxonomy to a .txt file, the parent order is random and not deterministic
To solve this, we are now keeping track of the parent order in the original file
In the future, we may want to impose the parent order (alphabetical order for example) but right now, it would result in too big .txt file diff

### Fixes bug(s)
https://github.com/orgs/openfoodfacts/projects/103/views/1?pane=issue&itemId=48206861

### DIFF
--- a/parser/openfoodfacts_taxonomy_parser/parser/parser.py
+++ b/parser/openfoodfacts_taxonomy_parser/parser/parser.py
@@ -151,17 +151,21 @@ class Parser:
                 unnormalised_child_links.append(child_link)
 
         # adding normalised links is easy as we can directly match parent entries
-        normalised_query = f"""
-            UNWIND $normalised_child_links as child_link
-            MATCH (p:{project_label}) USING INDEX p:{project_label}(id)
-            WHERE p.id = child_link.parent_id
-            MATCH (c:{project_label}) USING INDEX c:{project_label}(id)
-            WHERE c.id = child_link.id
-            CREATE (c)-[relations:is_child_of]->(p)
-            WITH relations
-            UNWIND relations AS relation
-            RETURN COUNT(relation)
-        """
+        normalised_query = (
+            f"""
+                UNWIND $normalised_child_links as child_link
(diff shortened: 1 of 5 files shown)


parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -151,17 +151,21 @@ def _create_child_links(
...
-            RETURN COUNT(relation)
-        """
+        normalised_query = (
+            f"""
+                UNWIND $normalised_child_links as child_link
+                MATCH (p:{project_label}) USING INDEX p:{project_label}(id)
+                WHERE p.id = child_link.parent_id
+                MATCH (c:{project_label}) USING INDEX c:{project_label}(id)
+            """
+            + """
```
alexgarel, line 161: that's smart to separate the non interpolated part ;-)

parser/openfoodfacts_taxonomy_parser/unparser.py
```diff
@@ -33,8 +33,12 @@ def get_all_nodes(self, multi_label):
...
             )
             WHERE h.id="__header__" AND f.id="__footer__"
-            UNWIND nodes(path) AS n
-            RETURN n , [(n)-[:is_child_of]->(m) | m ]
+            WITH nodes(path) AS nodes, range(0, size(nodes(path))-1) AS indexes
+            UNWIND indexes AS index
+            WITH nodes[index] AS n, index
+            OPTIONAL MATCH (n)-[r:is_child_of]->(parent)
+            WITH n, r, parent ORDER BY index, r.position
+            RETURN n, collect(parent)
```
alexgarel, line 41: This is quite a complicated query and you should document why you do that (having parents in the right order) and maybe a bit of how you do it.

alexgarel (approved):
Cool ! Here you harvest the benefits of a well organized code :-)

Note: Please, add the requested comment before merging.

PR #338, opened 2024-01-17
feat: Improve node creation performance

eric-nguyen-cs:
### What
Small follow-up PR to https://github.com/openfoodfacts/taxonomy-editor/pull/318
By updating the node creation query, we speed by about x2 the node creation performance.
So for example, Categories are created in about 15s instead of 25s

### DIFF
--- a/parser/openfoodfacts_taxonomy_parser/parser/parser.py
+++ b/parser/openfoodfacts_taxonomy_parser/parser/parser.py
@@ -32,24 +32,28 @@ class Parser:
 
     def _create_other_node(self, tx: Transaction, node_data: NodeData, project_label: str):
         """Create a TEXT, SYNONYMS or STOPWORDS node"""
-        position_query = """
-            SET n.id = $id
-            SET n.preceding_lines = $preceding_lines
-            SET n.src_position = $src_position
-        """
         if node_data.get_node_type() == NodeType.TEXT:
-            id_query = f"CREATE (n:{project_label}:TEXT) \n"
+            type_label = "TEXT"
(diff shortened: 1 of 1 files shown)


parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -32,24 +32,28 @@ def _get_project_name(self, taxonomy_name: str, branch_name: str):
...
+        node_tags_queries = [f"{key} : ${key}" for key in node_data.tags]
+
+        base_properties_query = """
+            id: $id,
+            preceding_lines: $preceding_lines,
+            src_position: $src_position
+        """
 
-        query = id_query + entry_query + position_query
+        properties_query = ",\n".join([base_properties_query, *node_tags_queries])
```
alexgarel, line 52: First time I see the "star" pattern for lists ! :heart:

alexgarel (approved):
LGTM

**Note**: we could also batch the other_nodes creation, although there are only few, but it does not complicate the code so much !

eric-nguyen-cs:
> LGTM
> 
> **Note**: we could also batch the other_nodes creation, although there are only few, but it does not complicate the code so much !

Yes we could do what we did for ENTRY nodes for TEXT, SYNONYMS and STOPWORDS nodes but that is such a small performance gain that I believe that can focus on that much later (or as a part of another refactor)

eric-nguyen-cs:
linked to : https://github.com/openfoodfacts/taxonomy-editor/issues/289

PR #354, opened 2024-01-18
fix(frontend): Handle xx language for translations edition

Piv94165:
Enhance language display on the frontend for "xx":

Currently, when attempting to display a language, the frontend behavior is incomplete : the "xx" language is not displayed. 
To improve this, the goal is to ensure that the 'xx' translation is displayed in all situations as a translation section. This adjustment will enable user to add "xx" translations, which are universal words.

image

In the taxonomy file, we observe the entry: 'xx:E100'. This signifies that 'E100' is a universal identifier applicable to all languages. Therefore, it is essential to ensure it's displayed in the taxonomy editor to be able to modify it.

With this PR we have now :

image
the "All languages" section is displayed if : 
- "All languages" was selected as in the image below
- there are words in "xx" language property for the node (so we can't hide the section)

The section if displayed is always at the second position, after the main language.

image

When 'All languages' is chosen, the language is locally stored as the preferred language. However, if it wasn't selected, and the section is visible due to the presence of universal words, 'All languages' is not stored locally.

Note : I've added changes after the comments for this [PR](https://github.com/openfoodfacts/taxonomy-editor/pull/304#pullrequestreview-1830020268)


alexgarel:
@Piv94165 I don't think we should  do it that way ! It is error prone, IMO.

We just should handle xx as a specific language name "All languages" (or something near).

The only specific thing could be that we always display this language if it exists.

Piv94165:
@alexgarel indeed it was not the right approach. So instead of gathering words from a language with "xx" words, I've added a section "All languages" as you propose.
I opted to showcase the section even in its emptiness as I believe it's valuable to introduce 'xx' words, even in the absence of xx property. Otherwise, user won't be able to do it in the editor. What do you think ?

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.tsx
```diff
@@ -95,8 +96,10 @@ export const ListTranslations = ({
 
   const shownLanguagesInfo = languagesToShow.map((languageCode: string) => {
     const languageName =
-      ISO6391.getName(languageCode) +
-      (languageCode === nodeObject.main_language ? " (main language)" : "");
+      languageCode === "xx"
+        ? "All languages"
+        : ISO6391.getName(languageCode) +
```
perierc, line 101:
```suggestion
      (languageCode === "xx"
        ? "All languages"
        : ISO6391.getName(languageCode)) +
```

I think the ternary should be between parentheses, so we display "All languages (main language)" if xx is the main language

taxonomy-editor-frontend/src/pages/editentry/LanguageSelectionDialog.tsx
```diff
@@ -52,15 +54,25 @@ const LanguageSelectionDialog = ({
...
+              .concat(ISO6391.getAllNames())
               .sort()
               .map((languageNameItem) => {
-                const languageCodeItem = ISO6391.getCode(languageNameItem);
+                const languageCodeItem =
+                  languageNameItem === "All languages"
+                    ? "xx"
+                    : ISO6391.getCode(languageNameItem);
                 return languageCodeItem === mainLanguageCode ? null : (
                   <MenuItem key={languageCodeItem} value={languageCodeItem}>
```
perierc, line 66: I feel like this `MenuItem` should also be `disabled` if the checkbox is disabled, because currently we can still click on the "All languages" option even if the checkbox looks disabled

taxonomy-editor-frontend/src/pages/editentry/LanguageSelectionDialog.tsx
```diff
@@ -52,15 +54,25 @@ const LanguageSelectionDialog = ({
               selected.map((langCode) => ISO6391.getName(langCode)).join(", ")
```
perierc, line 54: This has to be modified as well, because currently when "All languages" is selected, it is rendered as an empty string:
image

taxonomy-editor-frontend/src/pages/editentry/LanguageSelectionDialog.tsx
```diff
@@ -52,15 +54,25 @@ const LanguageSelectionDialog = ({
...
+            {["All languages"]
+              .concat(ISO6391.getAllNames())
               .sort()
               .map((languageNameItem) => {
-                const languageCodeItem = ISO6391.getCode(languageNameItem);
+                const languageCodeItem =
+                  languageNameItem === "All languages"
+                    ? "xx"
+                    : ISO6391.getCode(languageNameItem);
                 return languageCodeItem === mainLanguageCode ? null : (
```
perierc, line 65: Also could you please add the fact that the main language's `MenuItem` and `Checkbox` are disabled? It would be nice to have the same kind of behaviour 😁

perierc (changes requested):
Good, a few nits and UI issues though.

taxonomy-editor-frontend/src/pages/editentry/ListTranslations.tsx
perierc: I think this comment is still relevant 😅 (although I'm not sure xx can be a main language)

perierc:
Also a few thoughts about this feature:

- I'm thinking maybe the "All languages" section could be displayed always just below the main language? (which is not the case here if there was no xx entry at first and "All languages" was selected in the dialog)

- And maybe the "All languages" option could just be a checkbox in the dialog and not inside the languages `Select`, because it feels weird imo to mix this option with all the other languages sorted

Piv94165:
@perierc, I've carefully reviewed your suggestions and I concur with your points.

Placing "all languages" above the main language (below the main language header) would be more intuitive. This way, users will be aware that it is unnecessary to add 'xx' words in other sections.

Integrating "All languages" within the list alongside other languages might seem a bit awkward. Initially, I considered setting it as the first language in the list, but I decided against it to avoid resembling a dysfunctional "select all" checkbox. Your suggestion of having a checkbox outside this list is a good alternative.

@alexgarel, what are your thoughts on this? Do you believe it's worthwhile to invest time in making this second change?

eric-nguyen-cs (commented):
I'll let @perierc review the `LanguagesSelectionDialog.tsx` and `ListTranslations.tsx` files

alexgarel:
No review but I tested, it works very well :-)

perierc (commented):
Good job! A few nits

taxonomy-editor-frontend/src/pages/root-nodes/index.tsx
```diff
@@ -49,13 +53,53 @@ const RootNodes = ({
...
+      if (!response.ok) {
+        throw new Error("Failed to fetch project info");
+      }
+      return response.json();
+    },
+    refetchInterval: (d) => {
+      return d.state.status === "success" &&
+        d.state.data?.status === ProjectStatus.LOADING
+        ? 1000
+        : undefined;
```
perierc, line 77: I think it should be returning `false` according to the documentation

taxonomy-editor-frontend/src/pages/root-nodes/index.tsx
```diff
@@ -49,13 +53,53 @@ const RootNodes = ({
...
     data: nodes,
     isPending,
     isError,
-    errorMessage,
-  } = useFetch<RootEntriesAPIResponse>(`${baseUrl}rootentries`);
+    error,
+  } = useQuery<RootEntriesAPIResponse>({
+    queryKey: [rootNodesUrl],
+    queryFn: async () => {
+      console.log("finally fetched !");
```
perierc, line 89:
```suggestion
```

taxonomy-editor-frontend/src/pages/root-nodes/index.tsx
```diff
@@ -49,13 +53,53 @@ const RootNodes = ({
...
+      const response = await fetch(rootNodesUrl);
+      if (!response.ok) {
+        throw new Error("Failed to fetch project info");
```
perierc, line 92:
```suggestion
        throw new Error("Failed to fetch root nodes");
```

perierc (commented):
👀

perierc:
(not directly related to the work here)
I've noticed a weird behaviour: when a project is created from the frontend and the user is redirected to the root nodes page, the first request to the `/project` endpoint is "pending" until the end of the parsing; we should investigate that (not necessarily in this PR)

backend/editor/entries.py
```diff
@@ -286,6 +286,57 @@ def is_valid_branch_name(self):
...
+            MATCH (p:PROJECT {id: $project_name})
+            RETURN p AS info
+        """
+        params = {"project_name": self.project_name}
+
+        result = await get_current_transaction().run(query, params)  # returns a coroutine
+        info_record = await result.single()
+        info = info_record["info"] if info_record else None
+
+        return info
```
eric-nguyen-cs, line 325: There is also a `get_project_info` controller in `project_controller.py` you will have to modify it to use it in `api.py`
Message me if you want to update the call in `api.py` and delete this function in `entries.py`

backend/editor/entries.py
```diff
@@ -286,6 +286,57 @@ def is_valid_branch_name(self):
...
+        Helper function to update a Taxonomy Editor project status
+        """
+        query = """
+            MATCH (n:PROJECT)
+            WHERE n.id = $project_name
+            SET n.status = $status
+        """
+        params = {"project_name": self.project_name, "status": status}
+        with session.begin_transaction() as tx:
+            tx.run(query, params)
```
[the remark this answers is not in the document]
Piv94165: set_project_status is replaced by edit_project, that's right ?

taxonomy-editor-frontend/src/pages/root-nodes/index.tsx
Piv94165: you're right it's number or false

backend/editor/controllers/project_controller.py
```diff
@@ -13,7 +15,10 @@ async def get_project(project_id: str) -> Project:
     """
     params = {"project_id": project_id}
     result = await get_current_transaction().run(query, params)
-    return Project(**(await result.single())["p"])
+    project = (await result.single())["p"]
+    if project is None:
+        raise HTTPException(status_code=404, detail="Project not found")
+    return Project(**project)
```
perierc, line 21: @eric-nguyen-cs If the `project_id` does not exist in the database, this will cause a `TypeError: 'NoneType' object is not subscriptable` error because we try to access a property of a `None` object.

```suggestion
    project = await result.single()
    if project is None:
        raise HTTPException(status_code=404, detail="Project not found")
    return Project(**project["p"])
```

perierc (commented):
Good, 1 error remaining!

perierc (approved):

backend/editor/controllers/project_controller.py
eric-nguyen-cs: Thanks!

PR #362, opened 2024-01-25
fix(frontend): Enhance Edition of Properties for a Node

Piv94165:
### What
1. When adding a property and confirming the row, the new property is not instantly visible in the table.
2. After adding a property, confirming the row, and saving changes, the new property is visible in the table
3. When deleting a property and confirming, the row is not instantly removed from the table.
4. After deleting a property, confirming, and saving changes, the deleted property is not visible anymore
5. Editing properties is working correctly if we only edit the value, otherwise it doesn't work well

### Screenshot

https://github.com/openfoodfacts/taxonomy-editor/assets/106757110/e7ef1631-3c62-4e80-b24c-838ee8f9ccb5

### Fixes bug(s)

https://github.com/openfoodfacts/taxonomy-editor/assets/106757110/69f9dcba-9ec2-4f07-bca8-8fa413c280ed

I've changed the style of the table and removed pagination to help users to edit properties :
image

### DIFF
--- a/taxonomy-editor-frontend/src/pages/editentry/ListAllEntryProperties.jsx
+++ b/taxonomy-editor-frontend/src/pages/editentry/ListAllEntryProperties.jsx
@@ -99,14 +99,25 @@ const ListAllEntryProperties = ({ nodeObject, setNodeObject }) => {
               }),
             onRowDelete: (selectedRow) =>
               new Promise((resolve, reject) => {
-                // Delete property from rendered rows
-                const index = selectedRow.tableData.id;
-                const updatedRows = data.filter((obj) => !(index === obj.id));
-                setData(updatedRows);
+                try {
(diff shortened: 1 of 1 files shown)


alexgarel:
Works fine !

Still todo:
- https://github.com/openfoodfacts/taxonomy-editor/issues/379
- https://github.com/openfoodfacts/taxonomy-editor/issues/380

alexgarel (approved):
Works fine for me.

Although there are quirk, as we don't get appropriate error messages if the property name is not valid (we provoque a 500 error instead).
See #380 for that.

eric-nguyen-cs (commented):
Thanks for solving the bugs, it's much more usable!
However, I have a style/UX nit: could you make it so the UI does not "jump" when we do edits/save changes? (I think that leaving some vertical space below the table for the "Save changes" button to appear should do it)

PR #364, opened 2024-01-25
feat(parser,frontend,backend): conserve comments order and sort properties at export

### LINKED ISSUE #363: Conserve order and place of comments
Currently, all comments are parsed together for an entry node, and rewritten at the beginning of the node at export. We should keep the order of the comments during parsing to be able to rewrite them in the same order at export.

### LINKED ISSUE #365: Sort properties of entry nodes at export

perierc:
### What

- Currently, all comments are parsed together for an entry node, and rewritten at the beginning of the node at export. This PR modifies the parser to keep the order of the comments during parsing to be able to rewrite them in the same order at export.
For that, we create a comments field for each commented property/language in each entry node in Neo4J.
A backend controller (`update_node`) has also been updated to take these comments into account.
This PR also disables the display of comments on the frontend, to simplify the app for now.

- This PR also sorts the properties of an entry node at export to have a deterministic output.

### Part of 
Resolves #363 
Resolves #365


parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -196,6 +199,22 @@ def _remove_separating_line(self, data: NodeData) -> NodeData:
...
+            comments.append(data.comments_stack.pop()[1])
+            current_line -= 1
+        return comments[::-1]
+    
+    def _get_parent_comments(self, data: NodeData) -> list[str]:
+        """Get comments just above the current line"""
+        parent_comments = []
+        while data.preceding_lines and data.preceding_lines[-1] != "":
+            parent_comments.append(data.preceding_lines.pop())
+        return parent_comments[::-1]
```
eric-nguyen-cs, line 217: nit: In my opinion, the impact of impure functions in the code can be difficult to follow (even if that is the way most of the functions in this file are written)
To write a pure function for `_get_parent_comments` you could have
```python
def _update_node_data_with_parent_comments(self, data: NodeData) -> NodeData:
    # modifying the node_data and return the modified node_data
    return modified_node_data
```
(yes it may cause a slight overhead to create a new `dataclass` but we are writing an app in Python and not a game in C++, code readability is more important here) 
or
```python
def _get_parent_comments_and_preceding_lines(self, data: NodeData) -> tuple[list[str], list[str]]:
    # calculate the new parent_comments and preceding_line
    return new_parent_comments, new_preceding_lines
```
I personally prefer the first approach
Of course, these are not the only ways to write pure functions here, but the idea still stands

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -196,6 +199,22 @@ def _remove_separating_line(self, data: NodeData) -> NodeData:
...
         return data
+    
+    def _get_comments_above(self, data: NodeData, line_number: int) -> list[str]:
+        """Get comments just above the current line"""
+        comments = []
+        current_line = line_number - 1
+        while data.comments_stack and data.comments_stack[-1][0] == current_line:
+            comments.append(data.comments_stack.pop()[1])
+            current_line -= 1
+        return comments[::-1]
```
eric-nguyen-cs, line 210: nit: Same as `_get_parent_comments`

parser/openfoodfacts_taxonomy_parser/unparser.py
```diff
@@ -55,10 +55,10 @@ def get_tags_line(self, node, lc):
         return lc + ":" + line
 
     def list_property_and_lc(self, node):
-        """return a ordered list of properties with their language code (lc)"""
+        """return an ordered list of properties with their language code (lc)"""
         # there is no rule for the order of properties
         # properties will be arranged in alphabetical order
-        return [property[5:] for property in node if property.startswith("prop_")]
+        return sorted([property[5:] for property in node if property.startswith("prop_") and not property.endswith("_comments")])
```
eric-nguyen-cs, line 61: thanks!

backend/editor/entries.py
```diff
@@ -452,13 +452,23 @@ async def update_node(self, label, entry, new_node):
         curr_node = result[0]["n"]
         deleted_keys = set(curr_node.keys()) - set(new_node.keys())
 
+        # Check for deleted keys having associated comments and delete them too
+        for key in deleted_keys:
+            comments_key = key + "_comments"
+            if new_node.get(comments_key) is not None:
```
eric-nguyen-cs, line 458:
```suggestion
            if curr_node.get(comments_key) is not None:
```
Or am I mistaken?

backend/editor/entries.py
```diff
@@ -452,13 +452,23 @@ async def update_node(self, label, entry, new_node):
...
                 deleted_keys.add(key)
-                # Delete tags_ids if we delete tags of a language
-                if key.startswith("tags_") and "_ids_" not in key:
+                # Delete tags_ids and comments if we delete tags of a language
+                if key.startswith("tags_") and "_ids_" not in key and not key.endswith("_comments"):
                     deleted_keys.add("tags_ids_" + key.split("_", 1)[1])
+                    deleted_keys.add(key + "_comments")
+                # Delete prop comments if we delete prop
+                if key.startswith("prop_") and not key.endswith("_comments"):
+                    deleted_keys.add(key + "_comments")
```
eric-nguyen-cs, line 471: Is it not possible to simplify that just adding `deleted_key + "_comments"` to the `deleted_keys` set for each `deleted_key`?

eric-nguyen-cs (commented):
A few questions and nits but it seems mostly great!
NB: did you do a few checks on various taxonomy files? I'd bet that there are some very weird edges cases that we have not thought of

eric-nguyen-cs:
You can also add comment tests for the parser

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
alexgarel: @eric-nguyen-cs I agree that pure function would be better, but it's also good to stay consistent. So I would not change it now, just for this two functions (or do we already have only a few ?).

We could also simply add a TODO comment to the file to propose a rewrite with the pattern you proposed.

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -196,6 +199,22 @@ def _remove_separating_line(self, data: NodeData) -> NodeData:
...
+    
+    def _get_parent_comments(self, data: NodeData) -> list[str]:
+        """Get comments just above the current line"""
```
alexgarel, line 213: Maybe change the docstring not to be the same as _get_comments_above ;-)

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -226,14 +245,26 @@ def _harvest_entries(self, filename: str, entries_start_line: int) -> Iterator[N
                     self.parser_logger.error(msg)
                 else:
                     data = self._remove_separating_line(data)
+                    if data.get_node_type() == NodeType.ENTRY:
```
alexgarel, line 248: why only for NodeType.ENTRY ?

We also need to keep comments for stopwords and synonyms

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -317,7 +353,11 @@ def _harvest_entries(self, filename: str, entries_start_line: int) -> Iterator[N
...
                         if property_name:
-                            data.properties["prop_" + property_name + "_" + lc] = property_value
+                            prop_key = "prop_" + property_name + "_" + lc
+                            data.properties[prop_key] = property_value
+                            prop_comments = self._get_comments_above(data, line_number)
+                            if prop_comments:
+                                data.comments[prop_key + "_comments"] = prop_comments
 
         data.id = "__footer__"
         data.preceding_lines.pop(0)
```
alexgarel, line 363: don't we have to do something about comments there also (it's an edge case, it's ok if you just put a TODO comment).

alexgarel (commented):
Thanks !

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -162,12 +165,12 @@ def _header_harvest(self, filename: str) -> tuple[list[str], int]:
...
     def _entry_end(self, line: str, data: NodeData) -> bool:
         """Return True if the block ended"""
-        # stopwords and synonyms are one-liner, entries are separated by a blank line
-        if line.startswith("stopwords") or line.startswith("synonyms") or not line:
-            # can be the end of an block or just additional line separator,
-            # file_iter() always end with ''
-            if data.id:  # to be sure that it's an end
-                return True
+        if data.id.startswith("stopwords") or data.id.startswith("synonyms"):
+            # stopwords and synonyms are one-liners; if the id is set, it's the end
```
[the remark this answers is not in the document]
perierc: It's possible, but most of the time synonyms and stopwords are written one after the other with no blank lines between them (because they are one-liners so they don't need to be separated with blank lines), so you can't really know which line to assign comments to.
I think it's better to determine that comments should be written before and keep only the comments above, as this structures the file and makes it possible to rewrite the taxonomy identically on export, which is the aim here.

backend/editor/entries.py
perierc: I think it's better to use `new_node`, because if there was no comments before (i.e in `curr_node`) but some comments were added to `new_node`, using `curr_node.get` would not add the comments to the deleted keys, and they would then be added to the node in neo4j.

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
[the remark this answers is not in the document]
eric-nguyen-cs: Ok it makes sense thanks!

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
eric-nguyen-cs: I understand wanting to stay consistent although here I believe that making the 2 functions pure only makes the code more readable, without changing the readability of the rest of the code.
I just feel that writing more readable and maintainable code should be done incrementally

backend/editor/entries.py
eric-nguyen-cs: Yep, you are completely right!

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
perierc: For stopwords and synonyms, comments are stored in the `preceding_lines` property, as before

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
perierc: Comments are still stored in the `preceding_lines` property for the footer, no changes

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
alexgarel: ok

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
alexgarel: ok

alexgarel (approved):

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -199,22 +204,43 @@ def _remove_separating_line(self, data: NodeData) -> NodeData:
...
-    def _get_comments_above(self, data: NodeData, line_number: int) -> list[str]:
-        """Get comments just above the current line"""
-        comments = []
+
+    def _get_node_data_with_comments_above_key(
+        self, data: NodeData, line_number: int, key: str
+    ) -> NodeData:
+        """Returns the updated node data with comments above the given
+        key stored in the {key}_comments property."""
+        new_data = copy.deepcopy(data)
```
eric-nguyen-cs, line 213: Good!

eric-nguyen-cs (approved):
Good

PR #369, opened 2024-01-31
fix(backend,parser): remove stopwords from normalized tags on node update

### LINKED ISSUE #367: Remove stop words from normalised tags when node updated

perierc:
### What

This PR fixes the fact that the stopwords were not removed during normalization of the tags outside the parser (i.e during node update, or node creation).

- Added an optional `stopwords` parameter (dict) to the `normalize_text` utils function, in order to remove stopwords from a given text.
- Modified the `update_node` controller to remove the stopwords during normalization thanks to the new parameter
- Refactored the parser to make it use the `normalize_text` utils function instead of its own method defined in its class

### Part of 

Fixes #367

### DIFF
--- a/backend/editor/entries.py
+++ b/backend/editor/entries.py
@@ -60,13 +60,14 @@ class TaxonomyGraph:
         """
         params = {"id": entry}
         query = [f"""CREATE (n:{self.project_name}:{label})\n"""]
+        stopwords = await self.get_stopwords_dict()
 
         # Build all basic keys of a node
         if label == "ENTRY":
             # Normalizing new canonical tag
             language_code, canonical_tag = entry.split(":", 1)
             normalised_canonical_tag = parser_utils.normalize_text(
-                canonical_tag, main_language_code
+                canonical_tag, main_language_code, stopwords=stopwords
             )
 
             # Reconstructing and updation of node ID
@@ -438,6 +439,27 @@ class TaxonomyGraph:
         result = await get_current_transaction().run(query, {"id": entry})
         return await async_list(result)
 
+    async def get_stopwords_dict(self):
+        """
+        Helper function used for getting all stopwords in a taxonomy, in the form of a dictionary
+        where the keys are the language codes and the values are the stopwords in the corresponding language
+        """
+        query = f"""
+            MATCH (s:{self.project_name}:STOPWORDS) 
+            WITH keys(s) AS properties, s 
+            UNWIND properties AS property 
+            WITH s, property 
+            WHERE property STARTS WITH 'tags_ids' 
+            RETURN property AS tags_ids_lc, s[property] AS stopwords
+        """
+        result = await get_current_transaction().run(query)
+        records = await async_list(result)
+        stopwords_dict = {}
+        for record in records:
+            language_code = record["tags_ids_lc"].split("_")[-1]
+            stopwords_dict[language_code] = record["stopwords"]
+        return stopwords_dict
+
     async def update_node(self, label, entry, new_node):
         """
         Helper function used for updation of node with given id and label
@@ -471,6 +493,7 @@ class TaxonomyGraph:
 
         # Adding normalized tags ids corresponding to entry tags
         normalised_new_node = {}
+        stopwords = await self.get_stopwords_dict()
         for key in set(new_node.keys()) - deleted_keys:
             if key.startswith("tags_"):
                 if "_ids_" not in key:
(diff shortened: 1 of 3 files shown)


parser/openfoodfacts_taxonomy_parser/utils.py
```diff
@@ -4,7 +4,7 @@
 import unidecode
 
 
-def normalize_text(line: str, lang="default", char="-"):
+def normalize_text(line: str, lang="default", char="-", stopwords={}):
```
eric-nguyen-cs, line 7:
```suggestion
def normalize_text(line: str, lang: str="default", char: str="-", stopwords: dict[str, str] | None=None) -> str:
   if stopwords is None:
      stopwords={}
```
https://realpython.com/python-optional-arguments/#data-types-that-shouldnt-be-used-as-default-arguments

parser/openfoodfacts_taxonomy_parser/utils.py
```diff
@@ -29,6 +29,15 @@ def normalize_text(line: str, lang="default", char="-"):
...
     line = re.sub(r"-+", char, line)
     line = line.strip(char)
+
+    # Remove stopwords
+    if lang in stopwords:
+        stopwords = stopwords[lang]
+        line_surrounded_by_char = char + line + char
+        for stopword in stopwords:
+            line_surrounded_by_char = line_surrounded_by_char.replace(char + stopword + char, char)
+        line = line_surrounded_by_char[1:-1]
```
eric-nguyen-cs, line 39: Good!

backend/editor/entries.py
```diff
@@ -438,6 +439,28 @@ async def get_children(self, entry):
...
+            WITH s, property
+            WHERE property STARTS WITH 'tags_ids'
+            RETURN property AS tags_ids_lc, s[property] AS stopwords
+        """
+        result = await get_current_transaction().run(query)
+        records = await async_list(result)
+        stopwords_dict = {}
+        for record in records:
+            language_code = record["tags_ids_lc"].split("_")[-1]
+            stopwords_dict[language_code] = record["stopwords"]
```
eric-nguyen-cs, line 461: nit: perhaps map comprehension may be more readable? (I'm not sure)

backend/editor/entries.py
```diff
@@ -438,6 +439,28 @@ async def get_children(self, entry):
         result = await get_current_transaction().run(query, {"id": entry})
         return await async_list(result)
 
+    async def get_stopwords_dict(self):
```
eric-nguyen-cs, line 442:
```suggestion
    async def get_stopwords_dict(self) -> dict[str,str]:
```

backend/editor/entries.py
```diff
@@ -438,6 +439,28 @@ async def get_children(self, entry):
...
+            WHERE property STARTS WITH 'tags_ids'
+            RETURN property AS tags_ids_lc, s[property] AS stopwords
+        """
+        result = await get_current_transaction().run(query)
+        records = await async_list(result)
+        stopwords_dict = {}
+        for record in records:
+            language_code = record["tags_ids_lc"].split("_")[-1]
+            stopwords_dict[language_code] = record["stopwords"]
+        return stopwords_dict
```
eric-nguyen-cs, line 462: nit: I understand if you don't want to do this, but it would be cleaner and more readable to create a Pydantic model for the return type
(The best would be to have a simple Neo4J query and to have the data transformation step as part of a Pydantic model validator)

eric-nguyen-cs (approved):
LGTM
I approved but please ask for another look if you decide to add a Pydantic model

parser/openfoodfacts_taxonomy_parser/utils.py
alexgarel: Well spotted.

parser/openfoodfacts_taxonomy_parser/utils.py
alexgarel, line 39:
```bash
    replace_expr = "|".join(stopwords.get(lang, []))
    if replace_expr:
        # replace stopwords by "-"
        line = re.sub(r'\b(' + replace_expr + r')\b', '-', line)
```

alexgarel (approved):
Great :-)

backend/editor/entries.py
perierc: ` -> dict[str, list[str]]` but yes

backend/editor/entries.py
perierc: sorry I think I'll pass on that one 😅

parser/openfoodfacts_taxonomy_parser/utils.py
perierc: I've tried this and it doesn't seem to work because `-` are not considered as words

Issue #376, opened 2024-02-01
Remove languages and countries from editable taxonomies


Piv94165:
**Issue Description:**
Currently, there is an unintended capability to edit languages and countries taxonomies in the editor. This behavior needs correction.

**To Do:**
- [ ] Disable the ability to import languages and countries taxonomies from GitHub.
- [ ] Ensure that languages and countries taxonomies are not displayed in the existing projects table.

PR #377, opened 2024-02-01
fix(frontend): Remove Countries and Languages from editable taxonomies

Piv94165:
### What
Currently, there is an unintended capability to edit languages and countries taxonomies in the editor. This behavior needs correction.

### To Do:
- [x] Disable the ability to import languages and countries taxonomies from GitHub.
- [x] Ensure that languages and countries taxonomies are not displayed in the existing projects table.

### Part of 
[https://github.com/openfoodfacts/taxonomy-editor/issues/376](issue376)

### DIFF
--- a/taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
+++ b/taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
@@ -56,11 +56,6 @@ const GoToProject = ({ clearNavBarLinks }: Props) => {
               status: status,
             };
           }
-        )
-        .filter(
-          (project) =>
-            project.taxonomyName !== "Languages" &&
-            project.taxonomyName !== "Countries"
         );
 
       newProjects = backendProjects;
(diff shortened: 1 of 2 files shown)


taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
```diff
@@ -36,26 +36,32 @@ const GoToProject = ({ clearNavBarLinks }: Props) => {
...
+              errors_count: errors_count,
+              status: status,
+            };
+          }
+        )
+        .filter(
+          (project) =>
+            project.taxonomyName !== "Languages" &&
+            project.taxonomyName !== "Countries"
+        );
```
eric-nguyen-cs, line 64: I don't think this is necessary as we should reset the staging/prod dbs soon enough

alexgarel: Yes this is not production software yet !

alexgarel (approved):
Thanks ! 

I approve but you can remove the filter in index.tsx as proposed by @eric-nguyen-cs

PR #378, opened 2024-02-01
fix(frontend): Handle no error case on errors page

Piv94165:
###Current situation
if the project has no error while parsing and we naviagte to errors page, this is what we see :

Image

###TO-DO
if no errors, the table and warning are not displayed.

image

### Part of 
[https://github.com/openfoodfacts/taxonomy-editor/issues/383](url)


alexgarel (approved):
Tested ok.

Thanks !

Issue #356, opened 2024-01-18
Fix parsing errors


eric-nguyen-cs:
Duplicate ID, reading errors, child link creation, property order
```[tasklist]
### Tasks
- [ ] https://github.com/openfoodfacts/taxonomy-editor/issues/372
- [x] property order: https://github.com/openfoodfacts/taxonomy-editor/pull/364
- [x] reading error: https://github.com/openfoodfacts/openfoodfacts-server/pull/9730
- [ ] missing child link - better log: https://github.com/openfoodfacts/taxonomy-editor/pull/381
- [ ] missing child links - fix `ingredients.txt` and `origins.txt`
```

eric-nguyen-cs:
To be done after fixing exporting to a PR, as this will facilitate debugging the parsing issues

PR #381, opened 2024-02-01
chore: Log more expressive errors for missing child links

eric-nguyen-cs:
### What

- Taxonomy parser now checks for valid child links and logs out the missing child links in the parsing errors

### Screenshot
Screenshot 2024-02-01 at 17 45 58

### Fixes bug(s)
- Partly fixes: https://github.com/openfoodfacts/taxonomy-editor/issues/356

### DIFF
--- a/parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
+++ b/parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
@@ -361,7 +361,7 @@ class TaxonomyParser:
         for lc, child_link in missing_child_links:
             self.parser_logger.error(
                 f"Missing child link at line {child_link['line_position']}: "
-                f"parent id {child_link['parent_id']} not found in tags_ids_{lc}"
+                f"parent_id {child_link['parent_id']} not found in tags_ids_{lc}"
             )
             missing_child_links_positions.add(child_link["line_position"])
 
--- a/parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
+++ b/parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
@@ -326,50 +326,69 @@ class TaxonomyParser:
         data.src_position = line_number + 1 - len(data.preceding_lines)
         yield data
 
-    def _get_valid_child_links(self, entry_nodes: list[NodeData], child_links: list[ChildLink]):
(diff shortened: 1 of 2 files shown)


parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -343,13 +390,21 @@ def _create_taxonomy(self, filename: str) -> Taxonomy:
...
+                for position, (parent, line_position) in enumerate(entry.parent_tag):
+                    child_links.append(
+                        ChildLink(
+                            parent_id=parent,
+                            id=entry.id,
+                            position=position,
+                            line_position=line_position,
+                        )
+                    )
+        valid_child_links = self._get_valid_child_links(entry_nodes, child_links)
```
alexgarel, line 402: Good idea to isolate that in a single method !

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -324,6 +326,51 @@ def _harvest_entries(self, filename: str, entries_start_line: int) -> Iterator[N
...
+            # we collect all the tags_ids in a certain language
+            tags_ids = set()
+            for node in entry_nodes:
+                if "tags_ids_" + lc in node.tags:
+                    tags_ids.update(node.tags["tags_ids_" + lc])
+
+            # we check if the parent_id exists in the tags_ids
+            for child_link in lc_child_links:
+                if child_link["parent_id"] not in tags_ids:
+                    missing_child_links.append((lc, child_link))
```
alexgarel, line 358: Very good implementation, but it seems of a pity that we do not make the most of this computation.

If, instead of a set, tags_ids is a dict mapping tags_ids to their canonical id, it still can be used to check for existence, but we can also retrieve the real parent at the same cost.

This will simplify (and improve performance on) parent links creation in the graph.

alexgarel (approved):
Great ! Just a suggestion of an improvement that we can do.

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
eric-nguyen-cs: Yes good idea!

alexgarel (approved):
Perfect !

PR #385, opened 2024-02-03
feat: Delete projects when their Github PR is merged

eric-nguyen-cs:
### What

- Add a scheduled task (once a day) to check if the exported project have their PR merged in Github, and if that is the case, then deletes it (deleting a project also now deletes its associated nodes and relationships)

### Fixes bug(s)
Fixes https://github.com/orgs/openfoodfacts/projects/103/views/1?pane=issue&itemId=51867390


backend/pyproject.toml
```diff
@@ -49,6 +49,7 @@ uvicorn = { extras = ["standard"], version = "^0.23.2" }
 neo4j = "^5.14.0"
 openfoodfacts_taxonomy_parser = { path = "../parser", develop = true }
 python-multipart = "^0.0.6"
+apscheduler = "^3.10.4"
```
alexgarel, line 52: Is there any reason not to use the existing integration: https://pypi.org/project/fastapi_scheduler/ ?

backend/editor/scheduler.py
```diff
@@ -0,0 +1,35 @@
...
+from .models.project_models import ProjectStatus
+
+log = logging.getLogger(__name__)
+
+
+async def delete_merged_projects():
+    async with TransactionCtx():
+        exported_projects = await get_projects_by_status(ProjectStatus.EXPORTED)
+        for project in exported_projects:
+            pr_number = project.github_pr_url and project.github_pr_url.split("/")[-1]
```
alexgarel, line 18: nit: The small optimization that's really not needed but exists :grin::
```suggestion
            pr_number = project.github_pr_url and project.github_pr_url.rsplit("/", 1)[-1]
```

backend/editor/scheduler.py
```diff
@@ -0,0 +1,35 @@
...
+            if not pr_number:
+                log.warning(f"PR number not found for project {project.id}")
+                continue
+            github_object = GithubOperations(project.taxonomy_name, project.branch_name)
+            if await github_object.is_pr_merged(int(pr_number)):
+                await delete_project(project.id)
```
alexgarel, line 24: As those methods can fail, I would add a try / except with log error, to avoid having one project blocking all other projects removal.

backend/editor/github_functions.py
```diff
@@ -135,3 +135,17 @@ async def create_pr(self, description) -> PullRequest:
...
+        """
+        try:
+            await self.connection.rest.pulls.async_check_if_merged(
+                *self.repo_info, pull_number=pr_number
+            )
+            return True
+        except RequestFailed as e:
+            if e.response.status_code == 404:
+                return False
+            raise e
```
alexgarel, line 151:
```suggestion
            # re-raise in case of unexpected status code
            raise e
```

backend/editor/github_functions.py
```diff
@@ -135,3 +135,17 @@ async def create_pr(self, description) -> PullRequest:
...
+        Check if a pull request is merged
+        """
+        try:
+            await self.connection.rest.pulls.async_check_if_merged(
+                *self.repo_info, pull_number=pr_number
+            )
+            return True
+        except RequestFailed as e:
+            if e.response.status_code == 404:
+                return False
```
alexgarel, line 150: Why False in this case ? Do you think it might means PR is not yet created ? Maybe a comment can help here.

backend/editor/controllers/node_controller.py
```diff
@@ -0,0 +1,9 @@
+from ..graph_db import get_current_transaction
+
+
+async def delete_project_nodes(project_id: str):
```
alexgarel, line 4: A comment to say it includes errors would be better (because I had to check how errors where reported in parser to verify we did not forgot them).

```suggestion
async def delete_project_nodes(project_id: str):
    """Remove all nodes for project.

    This includes entries, stopwords, synonyms and errors
    """
```

backend/editor/api.py
```diff
@@ -49,7 +51,16 @@
...
+
+# Setup FastAPI app lifespan
+@contextlib.asynccontextmanager
+async def app_lifespan(app: FastAPI):
+    async with graph_db.database_lifespan():
+        with scheduler_lifespan():
+            yield
+
+
+app = FastAPI(title="Open Food Facts Taxonomy Editor API", lifespan=app_lifespan)
```
alexgarel, line 63: That's a cool pattern !

Thanks for refactoring.

alexgarel (changes requested):
Really great.

The only blocker is the try/except in scheduler.

backend/pyproject.toml
eric-nguyen-cs: This library is not an official one from `tiangolo` and also seems to need a PostgresSQL DB connection

backend/editor/github_functions.py
eric-nguyen-cs: I feel comments should be reserved to logic that is difficult to understand (I'm not sure about commenting Python syntax 🤔)

backend/editor/scheduler.py
eric-nguyen-cs: I don't think the overall function should fail (unless there is a connection problem to Github) but I grouped the tasks into `asyncio.gather` to run them independantly

backend/editor/github_functions.py
eric-nguyen-cs: Screenshot 2024-02-06 at 19 54 28
I've bumped the `githubkit` dependency so that there is a link to the `github` docs, that details the status code return values

backend/editor/github_functions.py
alexgarel: the point in the comment is the "why", here it's to specify why we re-raise.

If I propose a comment it's because reading, I have to stop and think to understand.

alexgarel (approved):
Thank you ! 

Pleas, look at the open comments before merging.

PR #396, opened 2024-02-09
chore: Upgrade frontend tooling

eric-nguyen-cs:
### What
- Change Javascript files into Typescript files (the errors that are not straightforward to fix are ignored for now)
- Added CRACO to be able to configure Webpack and Eslint
- Added aliases to remove deeply-nested relative paths (Typescript and Webpack config)
- Added more relevant Eslint config and fixed errors


alexgarel (dismissed):
Impressive ! But I will let @perierc review ;-)

@nobeeakon by any chance are you also interested ?

perierc (approved):
Good work, thanks, it will improve the developer experience!
I would be really happy if you could make everything work without any `@ts-expect-error` though 😇 (but it's ok if you're not motivated for that 😁)

eric-nguyen-cs:
> Good work, thanks, it will improve the developer experience a lot! I would be really happy if you could make everything work without any `@ts-expect-error` though 😇 (but it's ok if you're not motivated for that 😁)

I've fixed a few other TS errors but I feel that the most important is to add Typescript as to not make any further type error: we can fix the remaining errors when we change the related code
Moreover, I believe that most of the remaining errors come from the fact that the data from the backend API is untyped, which we will be able to fix later!

PR #405, opened 2024-02-13
feat: add automatic client SDK generation

eric-nguyen-cs:
### What
- Add an automatic client SDK generator: it uses the fact that we are able to export an OpenAPI spec from FastAPI and that this spec is able to be parsed into a Typescript client

The goal is in the future to incrementally add `Pydantic` Models as return types to the path functions to serve as output validators and as documentation for the OpenAPI spec.
Thus, adding a new API call on the frontend should now involves:
- adding a Pydantic model to the return type of the API call
- using the client SDK to call the RPC instead of manually calling the API using `fetch` or `axios`. The API return types should also be imported from the SDK

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,6 @@ local_config_lint: ## Run on lint configuration files
 local_generate_sdk: ## Generate client SDK from OpenAPI spec
 	@echo "🍜 Generating client SDK from OpenAPI spec"
 	cd backend && make generate_spec
-	npm run lint
 	cd taxonomy-editor-frontend && npm run generate:api
 
 
@@ -117,8 +116,6 @@ dev: build up ## Build and run the project
 # dev tools #
 #-----------#
 
-
-# Generate client SDK from OpenAPI spec
 generate_sdk: ## Generate client SDK from OpenAPI spec
 	@echo "🍜 Generating client SDK from OpenAPI spec"
 	${DOCKER_COMPOSE} run --rm taxonomy_api python -m openapi.generate_openapi_spec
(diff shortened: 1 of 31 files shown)


Makefile
```diff
@@ -111,6 +118,13 @@ dev: build up ## Build and run the project
 #-----------#
 
 
+# Generate client SDK from OpenAPI spec
```
perierc, line 121: nit: it is already written in the next line 😁

Makefile
```diff
@@ -71,6 +71,13 @@ local_config_lint: ## Run on lint configuration files
 	npm run lint
 
 
+local_generate_sdk: ## Generate client SDK from OpenAPI spec
+	@echo "🍜 Generating client SDK from OpenAPI spec"
+	cd backend && make generate_spec
+	npm run lint
```
perierc, line 77: can't you lint during the `make generate_spec`? It would look cleaner

perierc (approved):
Good work, I approve! I have a question though

backend/editor/models/types/datetime.py
alexgarel, line 1: :+1:

backend/openapi/openapi.json
alexgarel, line 1: Do you really want to put it in the git ?

Why not generate it with fastapi instead ?

eric-nguyen-cs: 🤔 Well I found it was easier to implement this way (and it may also facilitate the debugging of the SDK generation)
If you want I can have a look at how to do the FastAPI server but I feel that it may be a bit complex (namely what happens if the server is already running)

Issue #395, opened 2024-02-08
Cannot create new child (button hidden) if no children in entry


perierc:
In an entry's page, we can add new children with the + button:

![Image](https://github.com/openfoodfacts/taxonomy-editor/assets/82757576/e849c144-b099-4e9e-b978-338428fdd0f5)

However, if there are no children currently, this button is hidden:

![Image](https://github.com/openfoodfacts/taxonomy-editor/assets/82757576/5c6109cc-a8d6-470d-8225-f427191e0cb3)

PR #406, opened 2024-02-14
fix(frontend): enable child creation even if no children in entry

Piv94165:
### What
In an entry's page, we can add new children with the + button:
Image

However, if there are no children currently, this button is hidden:

Image

### Fixes bug(s)
image
The button is displayed in all cases.

### Part of 
https://github.com/openfoodfacts/taxonomy-editor/issues/395


perierc (approved):

eric-nguyen-cs (approved):
LGTM! I've stopped another related issue concerning the node children if you want to tackle it in another PR. https://github.com/openfoodfacts/taxonomy-editor/issues/407

PR #408, opened 2024-02-14
feat(parser): handle duplicate ID errors during parsing

### LINKED ISSUE #372: handle duplicate id while parsing
### What
on duplicate ids:
- merge translations
- if same properties, keep the value of the last one

perierc:
### What

- This PR merges the duplicate nodes (having the same ID) during parsing.
- When a node with an already existing ID is harvested, we yield it, but we don't link it to any other node (no PreviousLink is related to the duplicate)
- Added a new step to remove duplicate child links (due to the fact that duplicate nodes are yielded)
- Added a step to remove duplicate entry nodes: translations are merged (union), properties are overwritten with the latest values found, comments are added.

### Screenshot

The new error message for duplicate IDs looks like this:

image

### Part of 
- Resolves #372


parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -108,15 +108,15 @@ def _file_iter(self, filename: str, start: int = 0) -> Iterator[tuple[int, str]]
                 line_count += 1
             yield line_count, ""  # to end the last entry if not ended
 
-    def _add_line(self, line: str) -> str:
+    def _normalize_entry_id(self, raw_id: str) -> str:
```
alexgarel, line 111: :+1: thanks for renaming things.

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -438,6 +444,54 @@ def _get_valid_child_links(self, entry_nodes: list[NodeData], raw_child_links: l
...
+        ids_to_nodes = dict()
+        for node in entry_nodes:
+            if node.id in ids_to_nodes:
+                first_node = ids_to_nodes[node.id]
+                for key, value in node.tags.items():
+                    if not key.startswith("tags_ids_"):
+                        # union of the tags
+                        first_node.tags[key] = list(
+                            dict.fromkeys(first_node.tags.get(key, []) + value)
+                        )
```
alexgarel, line 472: why not a set ?
```suggestion
                        first_node.tags[key] = list(
                            set(first_node.tags.get(key, []) + value)
                        )
```

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -438,6 +444,54 @@ def _get_valid_child_links(self, entry_nodes: list[NodeData], raw_child_links: l
...
+                            for tag in first_node.tags[key]
+                        ]
+                for key, value in node.properties.items():
+                    # overwrite the value if the property already exists in the first node
+                    first_node.properties[key] = value
+                for key, value in node.comments.items():
+                    # union of the comments
+                    first_node.comments[key] = list(
+                        dict.fromkeys(first_node.comments.get(key, []) + value)
+                    )
```
alexgarel, line 487:
```suggestion
                    first_node.comments[key] = list(
                        set(first_node.comments.get(key, []) + value)
                    )
```

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -438,6 +444,54 @@ def _get_valid_child_links(self, entry_nodes: list[NodeData], raw_child_links: l
...
+    def _remove_duplicate_child_links(self, child_links: list[ChildLink]) -> list[ChildLink]:
+        """Remove duplicate child links (i.e child links with the same parent_id and id)"""
+        unique_child_links = []
+        children_to_parents = collections.defaultdict(set)
+        for child_link in child_links:
+            parent_id, child_id = child_link["parent_id"], child_link["id"]
+            if parent_id not in children_to_parents[child_id]:
+                children_to_parents[child_id].add(parent_id)
+                unique_child_links.append(child_link)
+        return unique_child_links
```
alexgarel, line 456: FYI another way to do it:

```python
# classify by edges, overwriting if there is more than one value
edges_to_link = {
    (child_link["parent_id"], child_link["id"]): child_link
    for child_link in child_links
}
# return links
return list(values(edges_to_link))
    
```

alexgarel (commented):
Really cool. 

Some cosmetic changes proposed.

@eric-nguyen-cs do you remember a good taxonomy to test this ?

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
perierc: because we lose the order of the translations if we convert to a set!

perierc:
> Really cool.
> 
> Some cosmetic changes proposed.
> 
> @eric-nguyen-cs do you remember a good taxonomy to test this ?

FYI I've tried with the Ingredients taxonomy (with 39 duplicate id errors) and I've verified manually that the first 10 errors were correctly merged, comparing with the original file, and it was all good

eric-nguyen-cs:
A few examples of the end result:
Screenshot 2024-02-14 at 16 15 30

Screenshot 2024-02-14 at 16 19 03

eric-nguyen-cs (approved):
LGTM

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
alexgarel: Ok, then add a comment to explain you use it to keep ordering.

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
perierc: very aesthetic! 
However I'm not sure that overwriting is a good thing here because it could increase the diff. Indeed child links have a `position` property to keep the parent order at unparsing, so I think it's better to get the order of the parents that are common with other duplicates from the first node, because the new parents of the other duplicates will be added to the first node of the file during unparsing. This way there will only be new lines added to the parents of the first node, and no swap between the previous parents of the first node. I hope I'm clear

alexgarel: No problem, it was just for fun ;-)

PR #418, opened 2024-02-15
fix: wording in export

teolemon:
### What
- fix: wording in export


aadarsh-ram:
I think you need to run `make checks` once and commit again.

aadarsh-ram (approved):
LGTM

PR #429, opened 2024-03-06
feat(parser,backend,frontend): enable extended taxonomies

### LINKED ISSUE #414: Solve problem of taxonomies extended by other taxonomies
For packaging and traces this is not a problem, they are just virtual taxonomies.

So it's a problem only for origins and ingredients.

[The perl code](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Tags.pm#L1157):
```perl
	# For the origins taxonomy, include the countries taxonomy
	if ($tagtype eq "origins") {
		@files = ("countries", "origins");
	}

	# For the Open Food Facts ingredients taxonomy, concatenate additives, minerals, vitamins, nucleotides and other nutritional substances taxonomies
	elsif (($tagtype eq "ingredients") and (defined $options{product_type}) and ($options{product_type} eq "food")) {
		@files = (
			"additives_classes", "additives", "minerals", "vitamins",
			"nucleotides", "other_nutritional_substances", "ingredients"
		);
	}

	# Packaging
	elsif (($tagtype eq "packaging")) {
		@files = ("packaging_materials", "packaging_shapes", "packaging_recycling", "preservation");
	}

	# Traces - just a copy of allergens
	elsif ($tagtype eq "traces") {
		@files = ("allergens");
	}
```

Proposed solution:

* add a pragma in taxonomies to tell external taxonomies to load, or put it in the taxonomy editor code ?
* parse files as done in perl code, but marking external entries with a property external
* do not enable edition on those entries and signal which taxonomy they are from

perierc:
### What
- For certain taxonomies (like ingredients), we need to concatenate different taxonomies before handling them. This PR enables the parser to handle taxonomies that should be extended before edit.
- Here is an example of an exported PR for `ingredients`: https://github.com/perierc/openfoodfacts-server/pull/4/files
- External nodes are not modifiable from frontend (except that you can add children to them):

image

- Added a label to identify external nodes from frontend, in root nodes page and search page (made this quickly since it should be rebuilt with the new search and root nodes page)
image

### Part of 
- Resolves #414


backend/editor/entries.py
```diff
@@ -120,11 +121,26 @@ async def get_and_parse_taxonomy(self, uploadfile: UploadFile | None = None):
...
+                        "additives",
+                        "minerals",
+                        "vitamins",
+                        "nucleotides",
+                        "other_nutritional_substances",
+                    ):
+                        external_filepath = await self.get_github_taxonomy_file(
+                            tmpdir, external_taxonomy
+                        )
+                        other_filepaths.append(external_filepath)
```
alexgarel, line 152: Could we put the list of external taxonomies in a configuration file instead ?

Maybe for now in settings.py 
```python
EXTERNAL_TAXONOMIES = {
    "ingredients": [
        "additives_classes",
        "additives",
        "minerals",
        "vitamins",
        "nucleotides",
        "other_nutritional_substances",
    ],
}
```

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -83,7 +83,8 @@ def _create_entry_nodes(self, entry_nodes: list[NodeData], project_label: str):
             id: entry_node.id,
             preceding_lines: entry_node.preceding_lines,
             src_position: entry_node.src_position,
-            main_language: entry_node.main_language
+            main_language: entry_node.main_language,
+            is_external: entry_node.is_external
```
alexgarel, line 87: sorry for this change, but I think it would be important to remember which taxonomy an entry comes from.

So that if we display the message that an entry is external, the user knows where to edit it.

So either:
* keep is_external and add an external_taxonomy property
* change is_external to external: with this property being empty meaning it's internal.

It could also be leverage in search to find entries from a specific external taxonomy !

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -234,14 +235,37 @@ def _write_to_database(self, taxonomy: Taxonomy, taxonomy_name: str, branch_name
...
+    ):
         """Process the file"""
         start_time = timeit.default_timer()
 
         branch_name = normalize_text(branch_name, char="_")
         taxonomy_parser = TaxonomyParser()
         try:
-            taxonomy = taxonomy_parser.parse_file(filename, self.parser_logger)
+            if external_filenames is None:
+                taxonomy = taxonomy_parser.parse_file(main_filename, self.parser_logger)
```
alexgarel, line 252: it's always better to have a code, that first unify cases, and then process.

I would rather write this as:
```python
            external_entry_nodes  = []
            if external_filenames:
                for filename in external_filenames:
                ... 
            taxonomy = taxonomy_parser.parse_file(
                    main_filename, self.parser_logger, external_entry_nodes
                )
```

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -464,6 +466,10 @@ def _merge_duplicate_entry_nodes(self, entry_nodes: list[NodeData]) -> list[Node
         for node in entry_nodes:
             if node.id in ids_to_nodes:
                 first_node = ids_to_nodes[node.id]
+                # if we merge a node from the taxonomy with an external node,
+                # the merged node is not external anymore
+                if first_node.is_external and not node.is_external:
+                    first_node.is_external = False
```
alexgarel, line 472: :+1: great you though of this.

That said, I'm not sure we want to support it !

I would even think about doing the reverse: even if a node is present in the taxonomy but is also in an external one, it should not be edited.
Otherwise, on export we will silently copy information from the other taxonomy and create duplication silently. 

On the other hands, doing as I just propose would mean removing an entry from the taxonomy :thinking: 

So I'm not sure what to do…

I would rather not support it yet… could we imagine keeping node separate in this case ? As it is a potential problem to have twice same id, I would change the external id to <id>@<external taxonomy name> (eg: en:cochinine@additives.txt) and leave it like that for now…

taxonomy-editor-frontend/src/constants.ts
```diff
@@ -40,7 +40,6 @@ export const TAXONOMY_NAMES = [
   "Nova Groups",
   "Nucleotides",
   "Nutrients",
-  "Origins",
```
alexgarel, line 43: Add a comment above on why we removed origins (as it depends on countries)

taxonomy-editor-frontend/src/pages/project/editentry/ListAllEntryProperties.tsx
```diff
@@ -124,76 +126,86 @@ const ListAllEntryProperties = ({ nodeObject, setNodeObject }) => {
...
-                  // Add new property to nodeObject
-                  changePropertyData(
-                    normalizeNameToDb(updatedRow.propertyName),
-                    updatedRow.propertyValue
-                  );
-                  resolve();
+          editable={
+            isExternalNode
+              ? undefined
+              : {
```
alexgarel, line 132: you were lucky it was that simple :star_struck:

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
alexgarel, line 1: Don't we need to handle "is_before" links ?

I think it's ok, but just to be sure we though about it !

parser/tests/integration/test_parser_integration.py
alexgarel, line 1: would be great to add a small test with an external taxonomy, ain't it ?
(maybe a very small test_external1.txt test_external2.txt taxonomies with 3 entries each)

taxonomy-editor-frontend/src/pages/project/editentry/index.tsx
```diff
@@ -45,13 +62,22 @@ const EditEntry = ({ taxonomyName, branchName, id }: EditEntryProps) => {
...
+              sx={{ ml: 1, color: greyHexCode }}
+              onClick={() => setOpenDeleteDialog(true)}
+            >
+              <DeleteOutlineIcon />
+            </IconButton>
+          )}
         </Stack>
+        {isExternalNode && (
+          <CustomAlert
+            message="This node has been imported from another taxonomy to extend the current taxonomy. You can only add children from the current taxonomy to it."
```
alexgarel, line 76: Indicating were the nodes comes from would be great. See my comment about changing is_external to another property style.

alexgarel (changes requested):
Great work, also I think we missed something in the spec !

parser/openfoodfacts_taxonomy_parser/parser/parser.py
```diff
@@ -234,14 +235,37 @@ def _write_to_database(self, taxonomy: Taxonomy, taxonomy_name: str, branch_name
...
+                    external_taxonomy = external_taxonomy_parser.parse_file(
+                        filename, self.parser_logger
+                    )
+                    external_entry_nodes.extend(external_taxonomy.entry_nodes)
+                for entry_node in external_entry_nodes:
+                    entry_node.is_external = True
+                # parse main taxonomy file
+                taxonomy = taxonomy_parser.parse_file(
+                    main_filename, self.parser_logger, external_entry_nodes
+                )
```
eric-nguyen-cs, line 267: What do you think if putting that as part of the `TaxonomyParser` (it would call itself recursively if there are external taxonomies)? 
The reason I am suggesting this is that I feel that we have two main concerns:
- parsing the `.txt` file into a Taxonomy AST
- transforming the Taxonomy AST into Cypher queries to be run to write the taxonomy to the database
And I feel that this code does a part of the Taxonomy AST creation

backend/editor/entries.py
```diff
@@ -126,11 +131,26 @@ async def get_and_parse_taxonomy(self, uploadfile: UploadFile | None = None):
...
-                    self.get_github_taxonomy_file(tmpdir)
+                    self.get_github_taxonomy_file(tmpdir, self.taxonomy_name)
                     if uploadfile is None
                     else self.get_local_taxonomy_file(tmpdir, uploadfile)
                 )
-                await run_in_threadpool(self.parse_taxonomy, filepath)
+                other_filepaths = None
+                if self.taxonomy_name == "ingredients":
+                    other_filepaths = []
+                    for external_taxonomy in (
```
eric-nguyen-cs, line 141: In terms of performance, you have here a waterfall of 7 different network calls. How much time does it take to load all the taxonomies? 
How difficult would it be to at least load the external taxonomies concurrently?

backend/editor/entries.py
eric-nguyen-cs: Would it be interesting to even put it at the top of the `ingredients.txt` file under a `external_taxonomies` entry for example?

taxonomy-editor-frontend/src/pages/project/editentry/index.tsx
```diff
@@ -23,6 +25,21 @@ const EditEntry = ({ taxonomyName, branchName, id }: EditEntryProps) => {
...
+  const url = createURL(taxonomyName, branchName, id);
+
+  const { data: node } = useQuery({
+    queryKey: [url],
+    queryFn: async () => {
+      const response = await fetch(url);
+      if (!response.ok) {
+        throw new Error("Failed to fetch node");
+      }
+      return response.json();
```
eric-nguyen-cs, line 37: Isn't it possible to use the `DefaultService` here?

taxonomy-editor-frontend/src/pages/project/editentry/index.tsx
```diff
@@ -45,13 +62,22 @@ const EditEntry = ({ taxonomyName, branchName, id }: EditEntryProps) => {
           <Typography sx={{ mb: 2, mt: 2, ml: 2 }} variant="h4">
             You are now editing &quot;{id}&quot;
```
eric-nguyen-cs, line 63: The text should depend on if the node is editable

taxonomy-editor-frontend/src/pages/project/editentry/index.tsx
eric-nguyen-cs: suggestion (probably to do a separate PR): if you keep track of the file SHA and with the `src_position`, you can even give a link to the Github file on the precise file (although it may not be true for huge taxonomies)

taxonomy-editor-frontend/src/pages/project/editentry/index.tsx
```diff
@@ -23,6 +25,21 @@ const EditEntry = ({ taxonomyName, branchName, id }: EditEntryProps) => {
...
+    queryFn: async () => {
+      const response = await fetch(url);
+      if (!response.ok) {
+        throw new Error("Failed to fetch node");
+      }
+      return response.json();
+    },
+  });
+
+  const isExternalNode = node?.is_external === true;
```
eric-nguyen-cs, line 41: nit: since you determine here if a node is external or not, I feel it may be clearer to just pass down a `isEditable` prop to the children (no need to access `node?.isExternal` anywhere else)

eric-nguyen-cs (commented):
Looks mostly good imo!

taxonomy-editor-frontend/src/pages/project/editentry/index.tsx
perierc: you can still "edit" by adding children to external nodes 🥸

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
perierc: The `is_before` links are used at export to get the order of the nodes of the edited taxonomy.
Here, I made sure that external nodes are not linked with an `is_before` to any other node, so that they do not appear on the exported file of the taxonomy.

backend/editor/entries.py
perierc: Good idea, I will launch the network calls concurrently

parser/tests/integration/test_parse_unparse_integration.py
```diff
@@ -60,37 +62,30 @@ def test_round_trip(neo4j):
...
             line = "<en:Passion fruit yogurts"
+        # third tweak: removing unexisting parents
+        elif line.startswith("<en:milk"):
+            continue
         expected_lines.append(line)
 
     assert expected_lines == lines
 
 
-def test_two_branch_round_trip(neo4j):
```
perierc, line 68: FYI: I removed this test because it was redundant with the test above and I didn't really see the point of it.

backend/editor/entries.py
alexgarel: Yep ! That would be a good idea too. We can make a PR to openfoodfacts-server, it's easy to accept such a PR.

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
alexgarel: perfect.

parser/tests/integration/test_parse_unparse_integration.py
alexgarel: The point was to verify we don't mess around when we have more than one project. (that we don't forget label in some part of the code).

So maybe we should instead keep this one.

alexgarel (approved):
Looks really good (my review was a bit quick because I lack time today, but important things are there, thank you)

backend/editor/entries.py
```diff
@@ -160,6 +149,25 @@ async def get_and_parse_taxonomy(self, uploadfile: UploadFile | None = None):
...
+        tasks = []
+
+        # Create tasks for each external taxonomy and store them in a list
+        for external_taxonomy in EXTERNAL_TAXONOMIES[self.taxonomy_name]:
+            task = asyncio.create_task(self.get_github_taxonomy_file(tmpdir, external_taxonomy))
+            tasks.append(task)
+
+        # Wait for all tasks to complete concurrently
+        for task in tasks:
+            external_filepath = await task
```
eric-nguyen-cs, line 166: big nit (feel free to ignore): isn't using something like `asyncio.gather` or `asyncio.TaskGroup` better here?
Also I believe you could use `anyio` directly which the concurrency library used by FastAPI under the hood  https://fastapi.tiangolo.com/async/#write-your-own-async-code

parser/openfoodfacts_taxonomy_parser/parser/taxonomy_parser.py
```diff
@@ -503,14 +509,24 @@ def _merge_duplicate_entry_nodes(self, entry_nodes: list[NodeData]) -> list[Node
...
+        self, filename: str, external_filenames: list[str] | None = None
     ) -> Taxonomy:
         """Create the taxonomy from the file"""
+        # parse external taxonomies if any, and add their entry nodes to the main taxonomy
+        external_entry_nodes = []
+        for external_filename in external_filenames or []:
+            external_taxonomy_parser = TaxonomyParser()
+            external_taxonomy = external_taxonomy_parser.parse_file(
+                external_filename, None, self.parser_logger
+            )
```
eric-nguyen-cs, line 521: I believe this should also work and may be cleaner
```suggestion
            external_taxonomy = self.parse_file(
                external_filename, None, self.parser_logger
            )
```

perierc: I don't think so, because of the `self.stopwords` dict that is constructed during `_harvest_entries`

PR #435, opened 2024-03-12
ci: restart policy must be always on deployment

alexgarel:
RESTART_POLICY must be set to always on deployments.


eric-nguyen-cs (approved):
Good catch!

perierc (approved):

PR #436, opened 2024-03-12
fix: Add camelCased aliases to Pydantic models

eric-nguyen-cs:
### What
Pydantic supports aliases: https://docs.pydantic.dev/latest/concepts/alias/
This is particularly useful to use camelCased variables on the frontend and the normal snake_cased variables in Python


backend/openapi/openapi.json
```diff
@@ -1248,36 +1245,38 @@
...
             "type": "string",
             "format": "date-time",
-            "title": "Created At"
+            "title": "Createdat"
           },
-          "github_checkout_commit_sha": {
+          "githubCheckoutCommitSha": {
             "anyOf": [{ "type": "string" }, { "type": "null" }],
-            "title": "Github Checkout Commit Sha"
+            "title": "Githubcheckoutcommitsha"
```
perierc, line 1260: a bit weird, but not really important

perierc (approved):
LGTM

backend/openapi/openapi.json
eric-nguyen-cs: Tracked in: https://github.com/openfoodfacts/taxonomy-editor/issues/437

Issue #442, opened 2024-03-13
QW: in project list, "EXPORTED" should be a link to the PR


eric-nguyen-cs:
instead of exported in the table, PR link

PR #443, opened 2024-03-13
feat: Update Project List

eric-nguyen-cs:
### What
- List Projects as anti-chronological order
- Add PR Link to Exported projects on the Project List
- Update the Project List style (namely you can just click the project's row to go to the project's root nodes instead of clicking a small edit action button)
- Refactor backend and frontend code (Pydantic model on the backend, better typing on the frontend, cleaner React code, better supported table component for MUI)

### Screenshot

https://github.com/openfoodfacts/taxonomy-editor/assets/77902715/745db888-e4d7-4c7a-835e-0844b18f1ecb

### Fixes bug(s)
Fixes https://github.com/openfoodfacts/taxonomy-editor/issues/441
Fixes https://github.com/openfoodfacts/taxonomy-editor/issues/442


backend/editor/entries.py
```diff
@@ -126,11 +126,20 @@ async def get_and_parse_taxonomy(self, uploadfile: UploadFile | None = None):
                 )
                 await run_in_threadpool(self.parse_taxonomy, filepath)
                 async with TransactionCtx():
-                    await edit_project(self.project_name, ProjectEdit(status=ProjectStatus.OPEN))
+                    error_node = await get_error_node(self.project_name)
+                    errors_count = len(error_node.errors) if error_node else 0
+                    await edit_project(
+                        self.project_name,
+                        ProjectEdit(status=ProjectStatus.OPEN, errors_count=errors_count),
```
alexgarel, line 133: Is this better than getting errors in a cipher query.

In the future I was imagining that appart parse errors, we would have logical errors that may change as you edit taxonomy.

alexgarel (commented):
All ok for me, I let @perierc or @Piv94165 test it !

backend/editor/entries.py
eric-nguyen-cs: I'm not sure: it seems that form the current use of errors, this seems to be cleaner and more performant
But if we add logical errors to the mix in the future that can be added when we edit a node, this may be not true..

alexgarel: It was only a question, I think we can leave it like it :-)

taxonomy-editor-frontend/src/pages/go-to-project/index.tsx
```diff
@@ -1,66 +1,89 @@
...
+      `/${toSnakeCase(params.row.taxonomyName)}/${params.row.branchName}/entry`
+    );
+  };
+
+  return (
+    <DataGrid
+      rows={projects}
+      columns={columns}
+      onRowClick={onRowClick}
+      pageSizeOptions={[25]}
```
perierc, line 75: This does not work: I have 35 projects and they all show in the same page.
I made it work doing this:
```suggestion
      pageSizeOptions={[]}
      initialState={{
        pagination: { paginationModel: { pageSize: 25 } },
      }}
```

perierc (approved):
Great work, but please fix the issue I've noticed before merging

PR #445, opened 2024-03-13
fix(backend): error on node creation from root nodes

### LINKED ISSUE #444: Error on creation of a node from root nodes page

perierc:
### What
- This PR fixes the bug that appears when the user tries to create a new node from the root nodes page.
- The issue was that for some reason, `NodeType.ENTRY` is not converted to its corresponding value (see error below). So I've decided for now to hard-code the "ENTRY" value, even if it's not optimal.

image

### Fixes bug(s)
- Fixes #444

### DIFF
--- a/backend/editor/api.py
+++ b/backend/editor/api.py
@@ -38,7 +38,7 @@ from .entries import TaxonomyGraph
 from .exceptions import GithubBranchExistsError, GithubUploadError
 
 # Data model imports
-from .models.node_models import EntryNodeCreate, ErrorNode, Footer, Header
+from .models.node_models import EntryNodeCreate, ErrorNode, Footer, Header, NodeType
 from .models.project_models import Project, ProjectEdit, ProjectStatus
 from .scheduler import scheduler_lifespan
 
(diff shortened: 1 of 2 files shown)


backend/editor/api.py
```diff
@@ -421,7 +421,7 @@ async def create_entry_node(
     normalized_id = await taxonomy.create_entry_node(
         new_entry_node.name, new_entry_node.main_language_code
     )
-    await taxonomy.add_node_to_end(NodeType.ENTRY, normalized_id)
+    await taxonomy.add_node_to_end("ENTRY", normalized_id)
```
alexgarel, line 424: you have to use .value
```suggestion
    await taxonomy.add_node_to_end(NodeType.ENTRY.value, normalized_id)
```

alexgarel: I let you re-introduce the import @perierc

alexgarel: BTW, to experiment: 
```python
docker compose run --rm taxonomy_api python3
>>> from editor.models.node_models import NodeType
>>> str(NodeType.ENTRY)
'NodeType.ENTRY'
>>> NodeType.ENTRY.value
'ENTRY'
```

perierc: Thanks for this solution and for the tip to experiment!
Finally I found that using a `StrEnum` makes it work also, so I went with that

alexgarel (approved):

PR #447, opened 2024-03-14
fix: error when adding child to node with no children

### LINKED ISSUE #446: Cannot add new child when there is no children currently

perierc:
### What
- This PR fixes the impossibility to add a child to a node that has currently no children.
- The error was due to a wrong Neo4J query in the `update_node_children` function in the backend.

### Fixes bug(s)
- Fixes #446


alexgarel (approved):
Great !

Good mastering of cipher :-)

PR #452, opened 2024-03-20
feat: add property filter to search API

eric-nguyen-cs:
### What
- Add the property filter to the search API
- The format is  `property:filter_value` where the filter value is in the format `not:inherited:property_name:lc:property_value`
   `not:` and `inherited:` and `:property_value` are optional.


eric-nguyen-cs:
Closing for a newer [PR](https://github.com/openfoodfacts/taxonomy-editor/pull/456) without merge conflits

PR #455, opened 2024-03-21
fix: fix backwards compatibility for is:not:external search

eric-nguyen-cs:
### What

- Add backwards compatibility for is:not:external search (`is_external` is null on older projects)


perierc (approved):
👍

PR #456, opened 2024-03-21
feat: add property filter to search API

eric-nguyen-cs:
### What
- Add the property filter to the search API
- The format is  `property:filter_value` where the filter value is in the format `not:inherited:property_name:lc:property_value`
   `not:` and `inherited:` and `:property_value` are optional.

### DIFF
--- a/backend/editor/models/search_models.py
+++ b/backend/editor/models/search_models.py
@@ -1,8 +1,9 @@
+import re
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from typing import Annotated, Literal
 
-from pydantic import Field, StringConstraints, TypeAdapter, computed_field
+from pydantic import Field, StringConstraints, TypeAdapter, computed_field, field_validator
 
 from .base_models import BaseModel
 from .node_models import EntryNode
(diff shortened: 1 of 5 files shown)


backend/editor/models/search_models.py
```diff
@@ -115,6 +116,130 @@ def build_cypher_query(self, param_name: str) -> CypherQuery:
...
+        terms = filter_value.split(":")
+        return terms[0] + "_" + terms[1]
+
+    @computed_field
+    def property_value(self) -> str | None:
+        filter_value = self.get_parsed_filter_value(self.negated, self.inherited)
+        terms = filter_value.split(":")
+        return ":".join(terms[2:]) if len(terms) > 2 else None
+
+    def get_parsed_filter_value(self, negated=False, inherited=False):
```
alexgarel, line 168: nit: instead of this pattern I would have added a cached_property "terms" that would have returned a tuple (or dict) with not / inherited / name and value from filter_value and used it everywhere :-)

alexgarel (approved):
Thanks

backend/editor/models/search_models.py
eric-nguyen-cs: Tracked: https://github.com/openfoodfacts/taxonomy-editor/issues/470

Issue #459, opened 2024-03-22
[QW] only display message about changing first entry if lang was not empty


perierc:
When I enter the first word for a new language, I got:
“Changing the first translation will modify the display name for this language!” -> displayed even if there was no translation at all for that language

PR #461, opened 2024-03-22
feat(frontend): make language selection for translations more intuitive

### LINKED ISSUE #460: Adding/selecting a language is not intuitive
- add a toggle to show all translations of an entries: continue to show selected (shown) languages, and non-selected languages are in grey with an icon, and we can click on this icon to select the language

perierc:
### What
The PR modifies the translations section of the edit entry page:
- Changes "All languages" terminology to "Fallback translations" (#458)
- Adds an info alert if "en" (English) or "xx" (Fallback translations) is not the main language for an entry (#457)
- Fixes the fact that the alert message about changing the display name of a language appears even if the language had no translations and we add the first one (#459)
- Adds a "Show all existing translations" checkbox to see all the languages that currently have translations, with their translations
- Adds a possibility to "pin" languages to select them (so they stay in local storage and appear at the top for each entry), and a possibility to hide (unselect) these languages easily (with an icon next to their title)
- Modifies the selection of new languages: I removed the "number of languages shown" button that had to be clicked to add a language, and created a "Show another language" button at the bottom of the section. Also, the dialog is now an autocomplete instead of a select, and you just type the languages that you want to add and see languages that are not selected, instead of seeing all current languages and being able to remove them. The autocomplete with the options is also automatically focused when opening the dialog.

### Screenshot

https://github.com/openfoodfacts/taxonomy-editor/assets/82757576/b9eadf88-1ade-467a-b947-536e9b8feac6

image

### Part of 
- Resolves #460 #459 #458 #457

### DIFF
--- a/taxonomy-editor-frontend/package-lock.json
+++ b/taxonomy-editor-frontend/package-lock.json
@@ -49,6 +49,7 @@
         "openapi-typescript-codegen": "^0.27.0",
         "prettier": "2.8.2",
         "vite": "^5.1.0",
+        "vite-plugin-svgr": "^4.2.0",
         "vite-tsconfig-paths": "^4.3.1"
       }
     },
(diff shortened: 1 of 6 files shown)


Piv94165 (approved):
Great PR!

Everything seems to be functioning smoothly from my end. One thing I might suggest is consulting with taxonomy experts regarding the naming convention for fallback languages, as Alizarine mentioned that she wasn't sure to understand the term. However, this can be addressed in a separate PR.

One minor issue I noticed is a peculiar behavior: when we display all existing languages, pin one from the list, and then hide it, it doesn't disappear (presumably because all existing languages are checked). Perhaps there's a better approach to handling this scenario. Nevertheless, I don't consider it critical enough to block the PR. 

Great work overall!

alexgarel:
Tested ok !

PR #466, opened 2024-03-27
fix(frontend): Replace white spaces in branch name proposition and allow dashes in owner name

Piv94165:
### What

There are some issues on the create project page:

- The owner's name allows for the inclusion of white spaces, while the branch name cannot accommodate them.
However, the proposed branch name includes the owner's name, potentially containing white spaces.
Therefore, to ensure compatibility, any white spaces within the branch name need to be replaced.

- While GitHub names can include dashes "-", the current text input field for the owner's name doesn't permit dashes, though it should.

### Screenshot

image


taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -32,9 +32,11 @@ export const StartProject = () => {
 
   const findDefaultBranchName = useCallback(() => {
     if (taxonomyName === "" || ownerName === "") return "";
-    return `${toSnakeCase(taxonomyName.toLowerCase())}_${ownerName
-      .replace(" ", "")
-      .toLowerCase()}_${Math.floor(Date.now() / 1000)}`;
+    return `${toSnakeCase(
+      taxonomyName.toLowerCase()
```
perierc, line 36: the `toSnakeCase` function already converts the argument to lower case, so the `.toLowerCase()` here is not needed

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -32,9 +32,11 @@ export const StartProject = () => {
...
+    )}_${ownerName}_${Math.floor(Date.now() / 1000)}`
```
perierc, line 37: It would be nice to make the date more readable for a human (maybe something like MMDDYY_HH_MinMin? for example for today: 032724_12_13)

perierc (approved):
It does the job, but maybe this could also be an opportunity to refactor other things around this default branch name 👀

alexgarel:
@perierc maybe we can merge this PR as currently it's more like a bug and do enhancement in a latter PR ?

alexgarel (commented):
Tested ok !

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -22,6 +22,19 @@ import { createBaseURL, toSnakeCase } from "@/utils";
 
 const branchNameRegEx = /[^a-z0-9_]+/;
 
+function formatDate(date) {
+  const map = {
+    mm: date.getMonth() + 1,
+    dd: date.getDate(),
+    yy: date.getFullYear().toString().slice(-2),
+    yyyy: date.getFullYear(),
```
perierc, line 30: nit: I think this `yyyy` variable is not used 👀

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -32,9 +45,11 @@ export const StartProject = () => {
 
   const findDefaultBranchName = useCallback(() => {
     if (taxonomyName === "" || ownerName === "") return "";
-    return `${toSnakeCase(taxonomyName.toLowerCase())}_${ownerName
-      .replace(" ", "")
-      .toLowerCase()}_${Math.floor(Date.now() / 1000)}`;
+    return `${taxonomyName.toLowerCase()}_${ownerName}_${formatDate(
```
perierc, line 48: nit: the `toLowerCase` for the taxonomy name is not needed because you use it again on the whole string after

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -22,6 +22,19 @@ import { createBaseURL, toSnakeCase } from "@/utils";
 
 const branchNameRegEx = /[^a-z0-9_]+/;
 
+function formatDate(date) {
+  const map = {
+    mm: date.getMonth() + 1,
+    dd: date.getDate(),
```
perierc, line 28: nit: I would make it so that these variables are always 2 characters long, even if the month or day is < 10 (I mean display 03 instead of 3 for March for instance, so that the format and number of characters is the same for all branches and so more readable)

perierc (commented):
a few nits 😅

taxonomy-editor-frontend/src/pages/startproject/index.tsx
```diff
@@ -22,6 +22,19 @@ import { createBaseURL, toSnakeCase } from "@/utils";
...
+    yy: date.getFullYear().toString().slice(-2),
+    yyyy: date.getFullYear(),
+    HH: date.getHours().toString().slice(-2),
+    MinMin: date.getMinutes().toString().slice(-2),
+  };
+
+  return `${map.mm}${map.dd}${map.yy}_${map.HH}_${map.MinMin}`;
+}
+
```
eric-nguyen-cs, line 37: Why not use the built-in `Date` methods? https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

eric-nguyen-cs (commented):
Another small comment

PR #468, opened 2024-03-27
fix(frontend): Prevent users from navigating to child with unsaved changes

Piv94165:
### What
Currently, users can navigate to a child node while editing a parent node without saving changes, resulting in a potential 404 error if the child node is newly created. 
Additionally, clicking on a child node before saving changes can lead to the loss of unsaved modifications. 

To prevent this, I have disabled the ability to navigate to a child node if this node has been newly created without saving. 
Furthermore, I have implemented a warning message to inform users about why they cannot proceed to the child node : "You've just created a new child. To navigate to it, please ensure your changes are saved first."

There is another warning massage when whatever changes have not been saved : "Changes are pending and have not been saved. Please save your changes before navigating to any child node. If you prefer not to save your pending changes but wish to avoid losing them, you can navigate to a child node in a new window."

What can be added :
- a button which enables user to discard changes

### Screenshot
The child node has been added whithout saving :
image

A change has not been saved :
image

### DIFF
--- a/taxonomy-editor-frontend/src/pages/project/editentry/ListEntryChildren.tsx
+++ b/taxonomy-editor-frontend/src/pages/project/editentry/ListEntryChildren.tsx
@@ -21,6 +21,7 @@ import CircularProgress from "@mui/material/CircularProgress";
 import ISO6391 from "iso-639-1";
 import { ENTER_KEYCODE } from "@/constants";
 import { greyHexCode } from "@/constants";
+import equal from "fast-deep-equal";
 
 interface Relations {
   index: string;
@@ -30,7 +31,10 @@ interface Relations {
 const ListEntryChildren = ({
   url,
   urlPrefix,
+  updateChildren,
   setUpdateNodeChildren,
+  previousUpdateChildren,
+  setPreviousUpdateChildren,
   hasChanges,
 }) => {
   const [relations, setRelations] = useState<Relations[]>([]);
@@ -52,6 +56,9 @@ const ListEntryChildren = ({
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-expect-error
       setUpdateNodeChildren(incomingData.map((el) => el?.[0]));
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-expect-error
+      setPreviousUpdateChildren(incomingData.map((el) => el?.[0]));
       const arrayData: Relations[] = [];
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-expect-error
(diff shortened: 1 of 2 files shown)


taxonomy-editor-frontend/src/pages/project/editentry/ListEntryChildren.tsx
```diff
@@ -122,6 +128,14 @@ const ListEntryChildren = ({ url, urlPrefix, setUpdateNodeChildren }) => {
         </IconButton>
       </Stack>
 
+      {/* Renders warning message to save changes to be able to click on a child node */}
+      {hasChanges ? (
```
perierc, line 132: nit: Since we don't show anything if there are no changes, I would rather write `hasChanges &&`

taxonomy-editor-frontend/src/pages/project/editentry/ListEntryChildren.tsx
```diff
@@ -122,6 +128,14 @@ const ListEntryChildren = ({ url, urlPrefix, setUpdateNodeChildren }) => {
...
+        <Alert severity="warning" sx={{ mb: 1, width: "fit-content" }}>
```
perierc, line 133: Could you please add a margin-left so that it is aligned with the "Children" title?

taxonomy-editor-frontend/src/pages/project/editentry/ListEntryChildren.tsx
```diff
@@ -131,7 +145,11 @@ const ListEntryChildren = ({ url, urlPrefix, setUpdateNodeChildren }) => {
             alignItems="center"
           >
             <Link
-              to={`${urlPrefix}/entry/${relationObject["child"]}`}
+              to={
+                hasChanges
+                  ? "#"
+                  : `${urlPrefix}/entry/${relationObject["child"]}`
+              }
```
perierc, line 152: The Alert is a wonderful idea, but I don't think we should change the link: imagine a user wants to open an existing child node in a new tab with a ctrl+click or right-click+open in new tab.

I think we should only remove the link for newly created children.

perierc (commented):
Good idea to add an Alert, but I don't think we should remove the href for all children (see comments)

taxonomy-editor-frontend/src/pages/project/editentry/ListEntryChildren.tsx
teolemon: Created: https://github.com/openfoodfacts/taxonomy-editor/issues/474

alexgarel (approved):
Great !

Tested on my side.

I stumbled on another bug while testing, see
- #485

PR #472, opened 2024-03-27
feat: add link to the Taxonomy Editor project in PR description

### LINKED ISSUE #471: QW: when we create a PR, the description should contain a link to the taxonomy editor project

perierc:
### What
- On PR creation via the editor, adds a link to the Taxonomy Editor project in the PR description.
- This introduces a `FRONTEND_URL` environment variable, that must be set for the link to appear correctly.

### Screenshot

image

### Part of 
- Resolves #471

### DIFF
--- a/backend/editor/github_functions.py
+++ b/backend/editor/github_functions.py
@@ -138,7 +138,7 @@ class GithubOperations:
         {description}
 
         ### Link to the project in Taxonomy Editor
-        {settings.frontend_url}/{self.taxonomy_name}/{self.branch_name}/entry
+        {settings.frontend_url}/{self.taxonomy_name}/{self.branch_name}
         """
         )
         return (
--- a/backend/editor/github_functions.py
+++ b/backend/editor/github_functions.py
@@ -137,6 +137,8 @@ class GithubOperations:
         ### Description
         {description}
 
+        ### Link to the project in Taxonomy Editor
+        {settings.frontend_url}/{self.taxonomy_name}/{self.branch_name}/entry
         """
         )
         return (
--- a/backend/editor/settings.py
+++ b/backend/editor/settings.py
@@ -6,6 +6,7 @@ import os
 uri = os.environ.get("NEO4J_URI", "bolt://localhost:7687")
 access_token = os.environ.get("GITHUB_PAT")
 repo_uri = os.environ.get("REPO_URI", "openfoodfacts/openfoodfacts-server")
+frontend_url = os.environ.get("FRONTEND_URL", "http://localhost:3000")
 
 EXTERNAL_TAXONOMIES = {
     "food_ingredients": [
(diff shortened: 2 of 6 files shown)


backend/editor/github_functions.py
```diff
@@ -137,6 +137,8 @@ async def create_pr(self, description) -> PullRequest:
         ### Description
         {description}
 
+        ### Link to the project in Taxonomy Editor
+        {settings.frontend_url}/{self.taxonomy_name}/{self.branch_name}/entry
```
eric-nguyen-cs, line 141: nit: I would feel better if you did that and added a redirection from `/` to `/entry` for project routes
```suggestion
        {settings.frontend_url}/{self.taxonomy_name}/{self.branch_name}
```

backend/editor/settings.py
```diff
@@ -6,6 +6,7 @@
 uri = os.environ.get("NEO4J_URI", "bolt://localhost:7687")
 access_token = os.environ.get("GITHUB_PAT")
 repo_uri = os.environ.get("REPO_URI", "openfoodfacts/openfoodfacts-server")
+frontend_url = os.environ.get("FRONTEND_URL", "http://localhost:3000")
```
eric-nguyen-cs, line 9:
```suggestion
frontend_url = os.environ.get("FRONTEND_URL", "YOUR_TAXONOMY_EDITOR_URL")
```
If the fallback is used, at least the users would be aware that they have to look for the correct URL (I basically do not like that forgetting to set this environment variable silently fails (same for the other env vars)

docker-compose.yml
```diff
@@ -35,6 +35,7 @@ services:
       # taken from .env
```
eric-nguyen-cs, line 35: Is there a reason why you have not added `FRONTEND_URL` to the `.env`?

eric-nguyen-cs (commented):
I'll let @alexgarel review this more in depth as it is his area of expertise

backend/editor/settings.py
alexgarel: Even: 
```suggestion
frontend_url = os.environ.get("FRONTEND_URL", "PLEASE_PROVIDE_FRONTEND_URL_ENV")
```

docker-compose.yml
alexgarel: I think it's because he was providing a fallback working for dev :-)

But I think it's better to do as @eric-nguyen-cs propose: put `FRONTEND_URL=http://ui.taxonomy.localhost:8091` in .env (this is the default when working with docker I think).

alexgarel (commented):
Good @perierc, you just have to fix comments raised by @eric-nguyen-cs

teolemon:
are we good to merge ? @eric-nguyen-cs @perierc

eric-nguyen-cs (approved):

PR #484, opened 2024-04-10
feat: improve search results display

### LINKED ISSUE #482: Advanced search: improve results display

perierc:
### What

- This PR improves the display of the search results
- It also fixes a bug if a node is created in the project, by removing the `src_position` from the EntryNode pydantic schema (since this property is not created when we create a new entry from the frontend, because it is not used)
- It also removes the old "Nodes" page and adds a redirection to the search page for the root and /entry paths of a project
- Also makes the text black in the top nav bar

### Screenshot

image

### Part of 
- Resolves #482


teolemon:
- Thanks for fixing my half-baked PR on coloring the top bar.

### Suggestions
- sorting languages in alphabetical order (or selection order, then alphabetical order)
- Renaming "External node" to something more intuitive

perierc:
@eric-nguyen-cs ready for review

PR #498, opened 2024-05-02
fix: sort languages in search filters

alexgarel:
From Alizarine: when I want to select a language in the "translated into" dropdown menu, the languages are not in the alphabetic order (see screenshot). It can be confusing but it's not a big deal.


alexgarel:
For failing black check, we have to first merge https://github.com/openfoodfacts/taxonomy-editor/pull/497

Issue #541, opened 2024-09-20
Change taxonomy generation to only change modified lines.


alexgarel:
Currently we are regenerating the whole taxonomy when we export it.

But this proved to be difficult to use. Taxonomies have to be perfect on Product Opener side to be easy to review and merge.

We need to switch strategy:

- keep whole taxonomy file content at import time
- register where each entry start / end (if it's not yet the case, but I think it's already done)
- [keep track of modified entries](https://github.com/openfoodfacts/taxonomy-editor/issues/366)
- at export time take original file and only replace the lines the entry was occupying with new lines corresponding to modified entry

alexgarel:
Done in #554

Issue #366, opened 2024-01-29
Add a flag to modified entries (with datetime value)


alexgarel:
1. Add a flag "modified" with the last modification date to an entry (non modified entry do not have it).
2. In search enable getting all modified entries
3. Eventually also add a sort on modification date

point 2 will be handy for reviewers of a taxonomy PR. The link to this view could be added to PR description.

point 3 might be handy for anyone editing a taxonomy to retrieve last modified entries

This is halfway before [tracking modifications fully](https://github.com/openfoodfacts/taxonomy-editor/issues/58).

PR #554, opened 2024-10-31
feat: taxonomy patch instead of re-generating

alexgarel:
The goal is to be able to patch taxonomy text files instead of re-generating them completely.

This will avoid having a lot of changes that are not related to the real modifications made by a contributor.

For this we need to
* add a "modified" property to entries (to track modified entries)
* track lines location of entries (to know where to change the original file)
* keeping removed entries (because we need to remove them from the original file)

Relates to: #541 and #366

TODO:
- [x] be able to patch taxonomy to generate the PR
- [x] add tests
- [x] add modified parameter in search
- [ ] ~add sort on modified parameter in search~ (sort is not done at all, for now)
- [x] test it locally (see https://github.com/openfoodfacts-bot/openfoodfacts-server/pull/63)
- [x] avoid repeating comments when replacing entry
- [x] put children after parents, not the other way around
- [x] on entry id change, also re-output children (especially the parent line)
- [x] avoid adding empty line at end of file (it will accumulate otherwise)
- [x] fix bug of simple entry creation, not being exported


alexgarel:
I still need to do some testing !

alexgarel:
I did a test https://github.com/openfoodfacts-bot/openfoodfacts-server/pull/61 it results there are still some bugs that I have to solve.

The most problematic is if we change entry id. I think that along the entry last modification timestamp, I will also store the modified attributes, so that I know if entry id was modified.

Issue #474, opened 2024-03-28
warning the user if he tries to close the page (or navigate away)


teolemon:
The loss of changes is however a big deal @perierc 

As we have hasChanges I think there is a way to handle that by warning the user if he tries to close the page (or navigate away):
see https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

It might be in another PR though.

_Originally posted by @alexgarel in https://github.com/openfoodfacts/taxonomy-editor/pull/468#discussion_r1541380003_

aapatil2004:
Can i work on this issue

alexgarel:
Just to be precise, as reported on the PR, navigate away means:
1. either closing the tab and / or typing / clicking a new external url
2. either navigating to another page of the app.

While the first might be solved with beforeunload_event,
The second needs to be handled in React routing (because the page does not really change from the browser point of view).

PR #555, opened 2024-12-14
Implemented warning on reload and exiting tab #474

aapatil2004:
This is how it looks when we user tries to reload filling partial information 
image
And this is how it looks when user tries to leave the site
leave

Implemented using Windows before unload feature

### DIFF
--- a/taxonomy-editor-frontend/package-lock.json
+++ b/taxonomy-editor-frontend/package-lock.json
@@ -29,7 +29,7 @@
         "web-vitals": "^3.5.0"
       },
       "devDependencies": {
-        "@craco/craco": "^7.1.0",
+        "@craco/craco": "^5.9.0",
         "@craco/types": "^7.1.0",
         "@testing-library/jest-dom": "^6.2.0",
         "@testing-library/react": "^14.2.1",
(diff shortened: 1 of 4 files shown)


alexgarel (commented):
Thanks @aapatil2004 for the good work. I have some remarks, that are more cleaning up, I imagine.

alexgarel (commented):
I have another problem with you implementation: it does not work if I click on a link in the application (click on the logo).

Is there a way to also handle it ?
getUserConfirmation (see [here](https://dev.to/eons/detect-page-refresh-tab-close-and-route-change-with-react-router-v5-3pd)) seems the way to go to handle this.

taxonomy-editor-frontend/src/components/ResponsiveAppBar.tsx
```diff
@@ -85,7 +100,7 @@ export const ResponsiveAppBar = () => {
                 page.url ? (
                   <MenuItem
                     key={page.translationKey}
-                    onClick={handleCloseNavMenu}
+                    onClick={() => handleNavigation(`/${page.url}`)} // Use handleNavigation
```
alexgarel, line 103: I think the method of using BrowserRouter getUserConfirmation proposed in [this article](https://dev.to/eons/detect-page-refresh-tab-close-and-route-change-with-react-router-v5-3pd) might be better (because more central). Did you try it ?

alexgarel (changes requested):
Thanks @aapatil2004 for your work, I have some issues though. I let you give a look at my comments.

teolemon:
@copilot can you fix the merge conflict ?

PR #562, opened 2025-01-15
build: Change docker commands for new version

aadarsh-ram:
### What
- The deploy workflow wasn't working due to commands being in the old Docker Compose version


alexgarel (approved):
Great to see you @aadarsh-ram :-) 

Thank you !

Issue #540, opened 2024-09-20
Do not account for stopwords as much as we did


alexgarel:
Currently we use stopwords to normalize entries, but this is wrong, Product Opener does not do this, it uses stop words only at matching time.

This leads taxonomy editor to think some entries are the same when they are not (eg. `fruit juice concentrate` vs `fruit juice from concentrate`) and may also rise synonyms positive that are not true positives. 

Re-evaluate at each point where we use stop-words if we need them or not (for parent matching we might try both, but raising error if there is more than one match with stop words removed).

PR #579, opened 2025-05-02
fix: do not account use stopwords for ids

alexgarel:
fixes: #540


stephanegigandet (approved):
Thank you!

aadarsh-ram:
@alexgarel Do any of the [sample JSON files](https://github.com/openfoodfacts/taxonomy-editor/tree/fix-stopwords/backend/sample) need changes? Is it still being used?

alexgarel:
> @alexgarel Do any of the [sample JSON files](https://github.com/openfoodfacts/taxonomy-editor/tree/fix-stopwords/backend/sample) need changes? Is it still being used?

I think we should remove them (in another PR maybe).

PR #696, opened 2026-07-12
ci: Add `/lint` command to lint the whole repository

Freso:
### What
- There’s currently a `/lint_code` comment command, but this only does `npm run lint` and doesn’t cover everything that `make lint` does.
- This adds a new `/lint` comment command to run `make lint` to lint everything, not just what `npm run lint` covers.


teolemon (approved):
That's great 🧹👌

Freso:
i'd've liked to test it, but it seems like it won’t want to run the command from the pr (which is probably a good call, for security reasons)

PR #725, opened 2026-08-31
fix: fix hishel version + avoid passing along uploadfile (just pass its content)

teolemon:
Disclaimer: code from Antigravity, Gemini Pro 3.1 with human review.

- fix hishel version to the one AsyncBaseStorage needs (githubkit will adapt)
- get uploaded file content before using background task, as it maybe closed when we read it otherwise.
- minor fixex in docker compose file (old version)


teolemon:
- What went wrong: The error ValueError: I/O operation on closed file. occurred because the /upload endpoint passed the incoming UploadFile object into a background task (background_tasks.add_task(self.get_and_parse_taxonomy, uploadfile)). FastAPI automatically closes uploaded files as soon as the HTTP route handler returns. By the time the background task actually ran, the file was already closed, causing the parsing step to fail.
- Because the taxonomy was never actually parsed and loaded into Neo4j (due to that error), the subsequent steps in the tests (like deleting a node or trying to export it) failed with empty results—leading to the downstream AttributeError: 'NoneType' object has no attribute 'split' and IndexError: list index out of range errors you saw.
- The Fix: I modified import_taxonomy, get_and_parse_taxonomy, and get_local_taxonomy_file in backend/editor/entries.py. Instead of passing the UploadFile object to the background task, the route handler now reads the file contents into memory as bytes (await uploadfile.read()) before returning, and then passes those raw bytes to the background task. This completely avoids the closed-file issue.

john-gom (approved):

backend/editor/entries.py
```diff
@@ -84,11 +83,10 @@ async def create_entry_node(self, name, main_language_code) -> str:
             EntryNodeCreate(name=name, main_language_code=main_language_code),
         )
 
-    async def get_local_taxonomy_file(self, tmpdir: str, uploadfile: UploadFile):
-        filename = uploadfile.filename
+    async def get_local_taxonomy_file(self, tmpdir: str, file_content: bytes, filename: str):
         filepath = f"{tmpdir}/{filename}"
         with open(filepath, "wb") as f:
-            await run_in_threadpool(shutil.copyfileobj, uploadfile.file, f)
+            f.write(file_content)
```
alexgarel, line 90:
```suggestion
            # it's not async but it should be fast enough (and rare)
            f.write(file_content)
```

alexgarel (approved):

